PR #10, opened 2019-09-16
Refactor KafkaWriter 

zzhhhzz:
Refactor KafkaWriter to attach headers that include hostname and locality information.

### DIFF
--- a/singer/src/main/java/com/pinterest/singer/writer/HeaderProvider.java
+++ b/singer/src/main/java/com/pinterest/singer/writer/HeaderProvider.java
@@ -1,0 +1,16 @@
+package com.pinterest.singer.writer;
+
+import com.pinterest.singer.common.SingerSettings;
+import org.apache.kafka.common.header.Headers;
+
+public class HeaderProvider {
+
+  public static final byte[] HOSTNAME = SingerSettings.getEnvironment().getHostname().getBytes();
+  public static final byte[] LOCALITY = SingerSettings.getEnvironment().getLocality().getBytes();
+
+  public static void addEnvHeaders(Headers headers){
+    headers.add("hostname", HOSTNAME);
+    headers.add("locality", LOCALITY);
+  }
+
+}
--- a/singer/src/main/java/com/pinterest/singer/writer/KafkaWriter.java
+++ b/singer/src/main/java/com/pinterest/singer/writer/KafkaWriter.java
@@ -31,6 +31,7 @@ import com.google.common.base.Strings;
 import org.apache.kafka.clients.producer.KafkaProducer;
 import org.apache.kafka.clients.producer.ProducerRecord;
 import org.apache.kafka.common.PartitionInfo;
+import org.apache.kafka.common.header.Headers;
 import org.apache.thrift.TSerializer;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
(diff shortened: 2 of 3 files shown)


singer/src/main/java/com/pinterest/singer/writer/HeaderProvider.java
```diff
@@ -0,0 +1,16 @@
...
+import com.pinterest.singer.common.SingerSettings;
+import org.apache.kafka.common.header.Headers;
+
+public class HeaderProvider {
+
+  public static final byte[] HOSTNAME = SingerSettings.getEnvironment().getHostname().getBytes();
+  public static final byte[] LOCALITY = SingerSettings.getEnvironment().getLocality().getBytes();
+
+  public static void addEnvHeaders(Headers headers){
+    headers.add("hostname", HOSTNAME);
```
HenryCaiHaiying: Define those string as constants.

singer/src/main/java/com/pinterest/singer/writer/HeaderProvider.java
```diff
@@ -0,0 +1,16 @@
+package com.pinterest.singer.writer;
+
+import com.pinterest.singer.common.SingerSettings;
+import org.apache.kafka.common.header.Headers;
+
+public class HeaderProvider {
+
+  public static final byte[] HOSTNAME = SingerSettings.getEnvironment().getHostname().getBytes();
```
HenryCaiHaiying: Get we also get Cell Id from SingerSettings?

singer/src/main/java/com/pinterest/singer/writer/KafkaWriter.java
```diff
@@ -221,19 +222,85 @@ private void includeAuditMessageInBatch(List<ProducerRecord<byte[], byte[]>> mes
...
     Set<Future<KafkaWritingTaskResult>> resultSet = new HashSet<>();
     KafkaProducer<byte[], byte[]> producer = KafkaProducerManager.getProducer(producerConfig);
     Preconditions.checkNotNull(producer);
-
     if (producerConfig.isTransactionEnabled()) {
       producer.beginTransaction();
     }
     try {
       List<PartitionInfo> partitions = producer.partitionsFor(topic);
+      /*
```
HenryCaiHaiying: Remove commented out code.

singer/src/main/java/com/pinterest/singer/writer/KafkaWriter.java
```diff
@@ -221,19 +222,85 @@ private void includeAuditMessageInBatch(List<ProducerRecord<byte[], byte[]>> mes
     return buckets;
   }
 
+  public List<List<LogMessage>> logMessagesCollation(List<PartitionInfo> partitions,
+                                                     String topic,
+                                                     List<LogMessage> logMessages) throws Exception {
+    LOG.info("Collate " + logMessages.size() + " messages");
```
HenryCaiHaiying: Would this be too much logging?

singer/src/main/java/com/pinterest/singer/writer/KafkaWriter.java
HenryCaiHaiying: Why do we need to break the previous methods into two methods?  Could header information be attached to the List<List<ProducerRecord>> generated from messageCollation() method?

singer/src/main/java/com/pinterest/singer/writer/KafkaWriter.java
ambud: Default log level is can control the volume of logging

zzhhhzz:
this PR is no longer needed.   newer implementation is https://github.com/pinterest/singer/pull/13

PR #12, opened 2019-09-27
add module audit-library (initial implementation) and AuditLogbackThriftLogger 

zzhhhzz:
The Singer package (Singer, thrift-logger) does not have a systematic way to auditing log message from end to end.  It will be beneficial to have logging audit capability to track each log message from end to end to see if log message get lost or what is the E2E latency. The data loss will be a  very important metric for any data pipeline.  I would like to add an audit-library module to Singer package and enable logging audit capability for ThriftLogger and Singer.  Also, this audit-library can be also imported and integrated by various possible systems of logging pipeline.  There would be several PRs to complete the implementation. 

This PR is about
1. refactor thrift-logger to move all thrift files to singer-commons module

2. add initial implementation of logging audit library to singer-commons

3. add AuditableLogbackThriftLogger, AuditableLogbackThriftLoggerFactory which can attach LoggingAuditHeaders to thrift object as well as to LogMessage.  A simple performance test shows that the adding headers will require extra 1 microsecond for each thrift message compared to using LogbackThriftLogger (no header added). The result is obtained by averaging 20 runs and each run generates 1 million thrift messages.  Test was done on EC2 c5.9xlarge instance with Ubuntu 14.04.5 and Java 8.


audit-library/src/main/java/com/pinterest/singer/loggingaudit/client/AuditHeadersGenerator.java
```diff
@@ -0,0 +1,52 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.pinterest.singer.loggingaudit.client;
+
+import com.pinterest.singer.loggingaudit.client.utils.AuditHeadersUtil;
+import com.pinterest.singer.loggingaudit.thrift.LoggingAuditHeaders;
+
+public class AuditHeadersGenerator {
```
yuyang08: add some comments to explain the motivation behind `AuditHeadersGenerator`?

audit-library/src/main/thrift/loggingaudit.thrift
```diff
@@ -0,0 +1,13 @@
+// Log message definitions written by thrift logger and pick up by singer agent.
+
+namespace py schemas.loggingaudit
+namespace java com.pinterest.singer.loggingaudit.thrift
+
+struct LoggingAuditHeaders {
+   1: required string host;         // source host that generates log message
+   2: required string topic;
```
yuyang08: add comments for each field?

thrift-logger/src/main/java/com/pinterest/singer/client/AuditLogbackThriftLogger.java
```diff
@@ -0,0 +1,163 @@
...
+  @Override
+  public void append(byte[] partitionKey, byte[] message, long timeNanos) {
+    throw new UnsupportedOperationException("For AuditLogbackThriftLogger, please use "
+        + "append(byte[] partitionKey, TBase thriftMessage, long timeNanos))");
+  }
+
+  @Override
+  public void append(byte[] partitionKey, TBase thriftMessage, long timeNanos) throws TException {
+    try {
+
```
yuyang08: format code and remove redundant empty lines?

thrift-logger/src/main/thrift/thrift_message.thrift
```diff
@@ -2,12 +2,15 @@
 
 namespace java com.pinterest.singer.thrift
 
+include "../../../../audit-library/src/main/thrift/loggingaudit.thrift"
```
yuyang08: refactor  modules in singer repo and move thrift related code to `singer-commons`?

thrift-logger/src/main/java/com/pinterest/singer/client/AuditLogbackThriftLogger.java
zzhhhzz: made the change in updated PR.

audit-library/src/main/thrift/loggingaudit.thrift
zzhhhzz: added comments in updated PR.

audit-library/src/main/java/com/pinterest/singer/loggingaudit/client/AuditHeadersGenerator.java
zzhhhzz: added comments in updated PR.

thrift-logger/src/main/thrift/thrift_message.thrift
zzhhhzz: The audit-library will be used by other components in the logging pipeline, for example Merced.  Thus, I feel it is better to all the code and thrift schema (LoggingAuditHeaders, AuditMessage, AuditConfig) in the audit-library module.

yuyang08: if we put it as part of `singer-commons`,  will it prevent Merced etc. from using it?  Merced can declare dependency on singer-commons.  Meanwhile `../../../../audit-library/src/main/thrift/loggingaudit.thrift` shows that the thrift files are not well organized.

zzhhhzz: I agree the thrift files importing does not looks good as a different module is added.  This will not prevent Merced and Sanitizer to use it. Eventually, the merced and Sanitizer will need 2 dependencies, one is singer-commons (including the all configurations classes for Singer), the other one is audit-library.   Initially I thought it is cleaner to have all audit-library related stuff in one module.  If you feel it is better to have singer-commons, we can go this approach.

yuyang08: shall we also consider audit-library as part of singer-commons?

zzhhhzz: I feel this is cleaner approach. I will make audit-library part of singer-commons and update the PR.

thrift-logger/src/main/java/com/pinterest/singer/client/AuditLogbackThriftLogger.java
```diff
@@ -0,0 +1,161 @@
...
+
+/**
+ * AuditLogbackThriftLogger is a class to log to Singer.
+ *
+ * Logback is used under the hood to log to and rotate the underlying
+ * file. The file is written using TFramedTransport as a format, so
+ * it is compatible with Singer's expected format.
+ */
+public class AuditLogbackThriftLogger extends BaseThriftLogger {
```
ambud: Rename to `Auditable`?

thrift-logger/src/main/java/com/pinterest/singer/client/AuditLogbackThriftLogger.java
```diff
@@ -0,0 +1,161 @@
...
+  }
+
+  @Override
+  public void append(byte[] message) {
+    throw new UnsupportedOperationException("For AuditLogbackThriftLogger, "
+        + "please use append(TBase thriftMessage))");
+  }
+
+  @Override
+  public void append(byte[] partitionKey, byte[] message, long timeNanos) {
```
ambud: Why are we not supporting this?

thrift-logger/src/main/java/com/pinterest/singer/client/AuditLogbackThriftLogger.java
```diff
@@ -0,0 +1,161 @@
...
+      OpenTsdbMetricConverter.incr(AUDIT_THRIFT_LOGGER_ERROR_INIT, "topic=" + topic,
+          "host=" + HOST_NAME);
+      LOG.error("Init method cannot finish successfully due to {}, needs to exit.", e.getMessage());
+      System.exit(1);
+    }
+  }
+
+  @Override
+  public void append(byte[] message) {
+    throw new UnsupportedOperationException("For AuditLogbackThriftLogger, "
```
ambud: Why are we not supporting this?

thrift-logger/src/main/java/com/pinterest/singer/client/AuditLogbackThriftLogger.java
zzhhhzz: The name AuditLogbackThriftLogger is following naming for LogbackThriftLogger. Both are concrete implementations by extending BaseThriftLogger. I feel AuditLogbackThriftLogger is better.

thrift-logger/src/main/java/com/pinterest/singer/client/AuditLogbackThriftLogger.java
zzhhhzz: We would like to inject LoggingAuditHeaders to the original thrift message, as its optional field.  If thrift message is passed as the argument, it is much efficient as we can set the LoggingAuditHeaders field directly; in the contrast, if message is passed as byte[], then we need to deserialize first.   If any applications use AuditLogbackThriftLogger, we want to enforce that the application is using **append(byte[] partitionKey, TBase message, long timeNanos)** rather than this one.

thrift-logger/src/main/java/com/pinterest/singer/client/AuditLogbackThriftLogger.java
zzhhhzz: Similar to above case.  We would like to inject LoggingAuditHeaders to the original thrift message, as its optional field.  If thrift message is passed as the argument, it is much efficient as we can set the LoggingAuditHeaders field directly; in the contrast, if message is passed as byte[], then we need to deserialize first.   If any applications use AuditLogbackThriftLogger, we want to enforce that the application is using **append(TBase message)** rather than this one.

thrift-logger/src/main/java/com/pinterest/singer/client/AuditLogbackThriftLogger.java
ambud: Audit is a verb in this case, Auditable is an adjective. Since classnames should be nouns we can use an adjective with it not a verb.

thrift-logger/src/main/java/com/pinterest/singer/client/AuditLogbackThriftLogger.java
ambud: The message bytes are wrapped in our custom thrift object so why does the payload matter?

thrift-logger/src/main/java/com/pinterest/singer/client/AuditLogbackThriftLogger.java
zzhhhzz: will change to AuditableLogbackThriftLogger in updated PR

thrift-logger/src/main/java/com/pinterest/singer/client/AuditLogbackThriftLogger.java
zzhhhzz: This method will be called by various applications that use thrift logger to log message to files.

thrift-logger/src/main/java/com/pinterest/singer/client/AuditLogbackThriftLogger.java
zzhhhzz: in the updated PR,  I am enabling append byte[]. This will give more flexibility.

PR #13, opened 2019-10-05
 Enables Singer to extract LoggingAuditHeaders from LogMessage and attach to the Headers of ProducerRecord

zzhhhzz:
This PR is about:

(1) add HeadersProvider interface and LoggingAuditHeadersProvider implementation;
(2) make headersProviderClass an option for SingerConfig (similar to environmentProviderClass);
(3) make enableHeadersProvider an option for SingerLogConfig to allow per topic configuration;
(4) modify KafkaWriter.java to attach LoggingAuditHeaders if configured properly.

### DIFF
--- a/singer/src/main/java/com/pinterest/singer/utils/LogConfigUtils.java
+++ b/singer/src/main/java/com/pinterest/singer/utils/LogConfigUtils.java
@@ -42,6 +42,8 @@ import com.pinterest.singer.thrift.configuration.TextLogMessageType;
 import com.pinterest.singer.thrift.configuration.TextReaderConfig;
 import com.pinterest.singer.thrift.configuration.ThriftReaderConfig;
 import com.pinterest.singer.thrift.configuration.WriterType;
+import com.pinterest.singer.writer.HeadersInjector;
+
 import com.google.common.base.Joiner;
 import com.google.common.base.Preconditions;
 import com.google.common.base.Strings;
@@ -348,6 +350,13 @@ public class LogConfigUtils {
     SingerLogConfig config = new SingerLogConfig(logName, local_dir, logfile_regex, processorConfig,
         readerConfig, writerConfig);
     config.setLogDecider(logDecider);
+    if (logConfiguration.containsKey("enableHeadersInjector")){
+      boolean enableHeadersInjector = logConfiguration.getBoolean("enableHeadersInjector");
+      config.setEnableHeadersInjector(enableHeadersInjector);
+      if (enableHeadersInjector && logConfiguration.containsKey("headersInjectorClass")){
+        config.setHeadersInjectorClass(logConfiguration.getString("headersInjectorClass"));
+      }
+    }
 
     FileNameMatchMode matchMode = FileNameMatchMode.PREFIX;
     String matchModeStr = logConfiguration.getString("logFileMatchMode");
@@ -365,6 +374,7 @@ public class LogConfigUtils {
       config
           .setLogRetentionInSeconds(logConfiguration.getInt(SingerConfigDef.LOG_RETENTION_SECONDS));
     }
+
     return config;
   }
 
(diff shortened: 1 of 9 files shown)


singer/src/main/java/com/pinterest/singer/writer/headersproviders/LoggingAuditHeadersProvider.java
```diff
@@ -0,0 +1,36 @@
...
+import org.apache.thrift.TSerializer;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class LoggingAuditHeadersProvider implements HeadersProvider {
+
+  private static final Logger LOG = LoggerFactory.getLogger(LoggingAuditHeadersProvider.class);
+
+  private static final String HEADER_KEY = "loggingAuditHeaders";
+  private static final TSerializer SER = new TSerializer();
```
ambud: @zzhhhzz unfortunately TSerializer is not thread safe (ref: https://issues.apache.org/jira/browse/THRIFT-2218) therefore this design will likely need to be changed. Since you are creating a singleton the serialize method will likely lead to data corruption.

singer/src/main/java/com/pinterest/singer/writer/KafkaWriter.java
```diff
@@ -212,6 +252,10 @@ private void includeAuditMessageInBatch(List<ProducerRecord<byte[], byte[]>> mes
           partitionId = validPartitions.get(partitionId).partition();
         }
         keyedMessage = new ProducerRecord<>(topic, partitionId, key, msg.getMessage());
+        if (enableHeadersProvider && SingerSettings.getHeadersProvider() != null){
```
ambud: Is it possible to & enableHeadersProvider with `SingerSettings.getHeadersProvider() != null` in constructor to save 1 condition evaluation?

singer/src/main/java/com/pinterest/singer/writer/headersproviders/LoggingAuditHeadersProvider.java
zzhhhzz: @ambud  thanks for pointing out this issue early.  Glad to know there is a patch in the jira ticket your mentioned,  but unfortunately it never get any updated.   I will update the PR to make  each Writer has one LoggingAuditHeaderProvider instance to prevent possible thread unsafe issues.

singer/src/main/java/com/pinterest/singer/writer/KafkaWriter.java
zzhhhzz: yes. it is possible. I will create LoggingAuditHeaderProvider instance per Writer to prevent the above issues and do the null check it in the constructor

singer/src/main/java/com/pinterest/singer/utils/LogConfigUtils.java
```diff
@@ -348,6 +350,13 @@ public static SingerLogConfig parseLogConfig(String logName,
     SingerLogConfig config = new SingerLogConfig(logName, local_dir, logfile_regex, processorConfig,
         readerConfig, writerConfig);
     config.setLogDecider(logDecider);
+    if (logConfiguration.containsKey("enableHeadersInjector")){
+      boolean enableHeadersInjector = logConfiguration.getBoolean("enableHeadersInjector");
+      config.setEnableHeadersInjector(enableHeadersInjector);
+      if (enableHeadersInjector && logConfiguration.containsKey("headersInjectorClass")){
+        config.setHeadersInjectorClass(logConfiguration.getString("headersInjectorClass"));
+      }
+    }
```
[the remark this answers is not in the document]
zzhhhzz: Using strings is to be consistent with other property parsing for SingerLogConfig.

PR #16, opened 2019-11-04
LoggingAudit Client library implementation

zzhhhzz:
(1) add LoggingAuditClient and its configuration parsing code. LoggingAuditClient maintains LinkedBlockingDeque of LoggingAuditEvent, has a AuditEventSender (wrapper of KafkaProducer) and provides audit() method for various clients. audit() method is called by client such as ThriftLogger instance and Singer's KafkaWriter in order to generate and enqueue a LoggingAuditEvent. LoggingAuditEvent is corresponding to LoggingAuditHeaders injected into the LogMessage.

(2) add implementations to singer module in order to enable LoggingAudit feature which is configurable per logStream (SingerLogConfig). SingerSettings will maintain an instance of LoggingAuditClient for handling LoggingAuditEvent generation and emission. Singer cleanup thread will call the close method of LoggingAuditClient to gracefully shutdown and cleanup LoggingAuditEvents left in the LinkedBlockingDeque.

(3) add implementations to thrift-logger module in order to enable LoggingAudit feature which is configurable per topic. AuditableLogbackThriftLoggerFactory will maintain an instance LoggingAuditClient for handling LoggingAuditEvent generation and emission. The close method of LoggingAuditClient is added at the end of shutdown method of AuditableLogbackThriftLoggerFactory.

(4) refactor ThriftLoggerFactory, AuditableLogbackThriftLoggerFactory and AuditableLogbackThriftLogger to make sure: AuditableLogbackThriftLoggerFactory is called by default in ThriftLoggerFactory. Whether LoggingAudit feature is enabled or not for AuditableLogbackThriftLoggerFactory depends on LoggingAuditClient being created properly; whether LoggingAudit feature is enabled or not for certain topic depends on ThriftLoggerConfig being created properly when creating AuditableLogbackThriftLogger instance. If LoggingAudit required fields are not set in ThriftLoggerConfig, the created AuditableLogbackThriftLogger instance behaves the same as the normal LogbackThriftLogger instance.


singer-commons/src/main/java/com/pinterest/singer/loggingaudit/client/AuditEventKafkaSender.java
```diff
@@ -0,0 +1,178 @@
...
+
+  public AuditEventKafkaSender(KafkaSenderConfig config,
+                               ArrayBlockingQueue<LoggingAuditEvent> queue,
+                               LoggingAuditStage stage , String host, String name){
+     this.topic = config.getTopic();
+     this.queue = queue;
+     this.stage = stage;
+     this.host = host;
+     this.name = name;
+     this.kafkaProducer = KafkaUtils.createKafkaProducer(config.getKafkaProducerConfig(), name);
```
ambud: Could we separate init and constructor to make this code testable?

singer-commons/src/main/java/com/pinterest/singer/loggingaudit/client/AuditEventKafkaSender.java
```diff
@@ -0,0 +1,178 @@
...
+               }
+             });
+           }catch(TException e){
+             LOG.warn("[{}] failed to construct ProducerRecord because of serialization exception.", Thread.currentThread().getName(), e);
+             OpenTsdbMetricConverter.incr(LoggingAuditClientMetrics.AUDIT_CLIENT_SENDER_SERIALIZATION_EXCEPTION, 1,
+                 "host=" +  host, "stage=" + stage.toString(), "logName=" + loggingAuditHeaders.getLogName());
+           }
+         }
+       }catch(InterruptedException e){
+         LOG.warn("[{}] got interrupted while polling the queue.", Thread.currentThread().getName(), e);
```
ambud: Interrupt exception doesn't quit out of the loop so in case of any other kind of thread termination this will be stuck. One possible design choice could be to use while(true) and then use an interrupt to simply quit out of the loop, this will avoid having to manipulate loop control to shutdown the thread.

singer-commons/src/main/java/com/pinterest/singer/loggingaudit/client/AuditEventKafkaSender.java
```diff
@@ -0,0 +1,178 @@
...
+  private AtomicBoolean cancelled = new AtomicBoolean(false);
+  private String topic;
+  private String name;
+  private Thread thread;
+  private int dequeueWaitInSeconds = 30;
+  private int stopGracePeriodInSeconds = 300;
+  private int callingThreadSleepInSeconds = 30;
+
+
+  public AuditEventKafkaSender(KafkaSenderConfig config,
```
ambud: Missing unit tests for this class.

thrift-logger/src/main/java/com/pinterest/singer/client/logback/AuditableLogbackThriftLoggerFactory.java
```diff
@@ -36,20 +49,86 @@
...
+    this.basePath = basePath;
+    this.rotateThresholdKBytes = rotateThresholdKBytes;
+  }
+
+  public AuditableLogbackThriftLoggerFactory() {
+    basePath = null;
+    rotateThresholdKBytes = -1;
+  }
+
+  public static void createLoggingAuditClientFromConfigFile(String filePath){
```
ambud: Seems like an unused method

ambud: Why are we using a static variable inside the factory? Shouldn't the factory itself be a singleton rather than the client?

thrift-logger/src/main/java/com/pinterest/singer/client/logback/AuditableLogbackThriftLoggerFactory.java
```diff
@@ -60,8 +139,38 @@ protected synchronized ThriftLogger createLogger(ThriftLoggerConfig thriftLogger
...
+        put("startAtCurrStage", "true");
+        put("stopAtCurrStage", "false");
+        put("samplingRate", String.valueOf(auditSamplingRate));
+      }
+    };
+  }
+
+  @Override
+  public synchronized void shutdown() {
+    for(String topic : loggersByTopic.keySet()){
```
ambud: Why not use an entryset iterator?

thrift-logger/src/main/java/com/pinterest/singer/client/logback/AuditableLogbackThriftLoggerFactory.java
```diff
@@ -36,20 +49,86 @@
...
+        rotateThresholdKBytes,
+        contextBase,
+        maxRetentionHours));
   }
 
   @Override
   protected synchronized ThriftLogger createLogger(ThriftLoggerConfig thriftLoggerConfig) {
     if (thriftLoggerConfig.baseDir == null || thriftLoggerConfig.logRotationThresholdBytes <= 0 ||
-        thriftLoggerConfig.kafkaTopic == null || thriftLoggerConfig.thriftClazz == null) {
+        thriftLoggerConfig.kafkaTopic == null ) {
```
ambud: Please access variables using getters and setters

singer/src/main/java/com/pinterest/singer/writer/KafkaWriter.java
```diff
@@ -338,7 +364,25 @@ public void writeLogMessages(List<LogMessage> logMessages) throws LogStreamWrite
...
+            for (Map.Entry<Integer, LoggingAuditHeaders> entry : listOfHeadersMap.get(bucketIndex)
+                .entrySet()) {
+                Integer listIndex = entry.getKey();
+                LoggingAuditHeaders loggingAuditHeaders = entry.getValue();
+                long messageAcknowledgedTimestamp = -1;
+
+                if (result.getRecordMetadataList().size() != buckets.get(bucketIndex).size()){
+                  LOG.error("Number of ProducerRecord does not match the number of RecordMetadata");
+                } else {
+                  RecordMetadata recordMetadata =result.getRecordMetadataList().get(listIndex);
```
ambud: Why does this need to be checked for every record? Isn't it sufficient to do this only once?

singer/src/main/java/com/pinterest/singer/writer/KafkaWriter.java
```diff
@@ -338,7 +364,25 @@ public void writeLogMessages(List<LogMessage> logMessages) throws LogStreamWrite
           // get the max write latency
           maxKafkaBatchWriteLatency = Math.max(maxKafkaBatchWriteLatency,
               result.getKafkaBatchWriteLatencyInMillis());
+
+          if (SingerSettings.loggingAuditClient != null) {
```
ambud: Please access variables using getters and setters

thrift-logger/src/main/java/com/pinterest/singer/client/logback/AuditableLogbackThriftLoggerFactory.java
```diff
@@ -60,8 +139,38 @@ protected synchronized ThriftLogger createLogger(ThriftLoggerConfig thriftLogger
         contextBase,
         thriftLoggerConfig.maxRetentionSecs / (60 * 60));    // lowest granularity is hours
 
+    if (loggingAuditClient != null && (thriftLoggerConfig.thriftClazz != null || thriftLoggerConfig.enableLoggingAudit)){
```
ambud: Please use getters and setters

thrift-logger/src/main/java/com/pinterest/singer/client/logback/AuditableLogbackThriftLoggerFactory.java
```diff
@@ -60,8 +139,38 @@ protected synchronized ThriftLogger createLogger(ThriftLoggerConfig thriftLogger
...
+      Map<String, String> properties = createProperties(thriftLoggerConfig.auditSamplingRate);
```
ambud: Getters and setters

singer/src/main/java/com/pinterest/singer/common/SingerSettings.java
```diff
@@ -94,6 +95,10 @@
   //environment is production by default
   private static Environment environment = new Environment();
 
+  // loggingAuditClient is used to send LoggingAuditEvent if LoggingAudit feature is enabled and
+  // a TopicAuditConfig is set for a given logStream.
+  public static LoggingAuditClient loggingAuditClient = null;
```
ambud: why is this variable public? please use getters and setters

thrift-logger/src/main/java/com/pinterest/singer/client/logback/AuditableLogbackThriftLoggerFactory.java
zzhhhzz: use entryset in the update of the PR.

thrift-logger/src/main/java/com/pinterest/singer/client/logback/AuditableLogbackThriftLoggerFactory.java
zzhhhzz: (1) the static method will be used in applications that uses ThriftLogger and can have audit configuration specified in a file.  This is just another way to create LoggingAuditClient and will help in migration from existing LogbackThriftLoggerFactory to AuditableLogbackThriftLoggerFactory.

(2) As a static variable of the factory, LoggingAuditClient is shared by all AuditableLogbackThriftLoggers created using the factory.  The factory itself is a singleton according the ThriftLoggerFactory.java.

singer-commons/src/main/java/com/pinterest/singer/loggingaudit/client/AuditEventKafkaSender.java
zzhhhzz: added tests in the update of the PR.

singer-commons/src/main/java/com/pinterest/singer/loggingaudit/client/AuditEventKafkaSender.java
zzhhhzz: refactored in the update of the PR.

singer-commons/src/main/java/com/pinterest/singer/loggingaudit/client/AuditEventKafkaSender.java
zzhhhzz: When thread is interrupted, then we just break the while loop and exits.  made the change in the update of the PR

singer-commons/src/main/java/com/pinterest/singer/loggingaudit/client/AuditEventKafkaSender.java
```diff
@@ -0,0 +1,186 @@
...
+  public KafkaProducer<byte[], byte[]> getKafkaProducer() {
+    return kafkaProducer;
+  }
+
+  public void setKafkaProducer(KafkaProducer<byte[], byte[]> kafkaProducer) {
+    this.kafkaProducer = kafkaProducer;
+  }
+
+  @Override
+  public void run() {
```
yuyang08: As reliable audit message writing is essential to guarantee auditing accuracy, we need to improve writing reliability by having multiple retries and randomly selecting different partitions during retry to minimize the impact of failure on a specific partition.

zzhhhzz: made changes in the update of the PR. also tested with integration test and performance test.

singer/src/main/java/com/pinterest/singer/common/SingerSettings.java
zzhhhzz: change to getter and setter

thrift-logger/src/main/java/com/pinterest/singer/client/logback/AuditableLogbackThriftLoggerFactory.java
zzhhhzz: made change

thrift-logger/src/main/java/com/pinterest/singer/client/logback/AuditableLogbackThriftLoggerFactory.java
zzhhhzz: made the change in update of the PR

singer/src/main/java/com/pinterest/singer/writer/KafkaWriter.java
zzhhhzz: made the change in update of the PR

singer/src/main/java/com/pinterest/singer/writer/KafkaWriter.java
zzhhhzz: No need. made change in the update of the PR.  only check once per partition (one bucket of buckets)

singer-commons/pom.xml
```diff
@@ -54,6 +54,26 @@
...
     </dependency>
+    <dependency>
+      <groupId>org.apache.kafka</groupId>
+      <artifactId>kafka_2.11</artifactId>
+      <version>2.1.0</version>
+    </dependency>
+    <dependency>
+      <groupId>org.apache.kafka</groupId>
+      <artifactId>kafka-clients</artifactId>
+      <version>2.1.0</version>
```
ambud: Refactor version to property in pom.xml ?

singer-commons/pom.xml
```diff
@@ -54,6 +54,26 @@
       <artifactId>commons-configuration</artifactId>
       <version>1.9</version>
     </dependency>
+    <dependency>
+      <groupId>org.apache.kafka</groupId>
+      <artifactId>kafka_2.11</artifactId>
+      <version>2.1.0</version>
```
ambud: Refactor version to property in pom.xml for reuse

singer-commons/src/main/java/com/pinterest/singer/loggingaudit/client/AuditEventKafkaSender.java
```diff
@@ -0,0 +1,250 @@
...
+    LoggingAuditEvent event = null;
+    ProducerRecord<byte[], byte[]> record;
+    byte[] value = null;
+    List<PartitionInfo> partitionInfoList = this.kafkaProducer.partitionsFor(topic);
+    LOG.info("Partitions of {}: {}", topic, partitionInfoList);
+    for(PartitionInfo partitionInfo : partitionInfoList){
+      partitionErrorCounts.put(partitionInfo.partition(), new AtomicInteger(0));
+    }
+     while(!cancelled.get()){
```
ambud: Any reason for still using a loop control?

singer-commons/src/main/java/com/pinterest/singer/loggingaudit/client/AuditEventKafkaSender.java
```diff
@@ -0,0 +1,250 @@
...
+  public KafkaProducer<byte[], byte[]> getKafkaProducer() {
+    return kafkaProducer;
+  }
+
+  public void setKafkaProducer(KafkaProducer<byte[], byte[]> kafkaProducer) {
+    this.kafkaProducer = kafkaProducer;
+  }
+
+
+  public int getGoodPartition(Set<Integer> badPartitions, int numOfPartitions){
```
ambud: `alternatePartition` or `randomPartition` ? This method doesn't check if a partition is good or bad.

singer-commons/src/main/java/com/pinterest/singer/loggingaudit/client/AuditEventKafkaSender.java
zzhhhzz: if exception happens,  we catch the exception and break the loop.

singer-commons/src/main/java/com/pinterest/singer/loggingaudit/client/AuditEventKafkaSender.java
zzhhhzz: Good means potentially good. we just want a different partition that is not tried before

singer-commons/pom.xml
zzhhhzz: made the change in the update of the PR.

singer-commons/src/main/java/com/pinterest/singer/loggingaudit/client/AuditEventKafkaSender.java
```diff
@@ -0,0 +1,340 @@
...
+import org.apache.kafka.clients.producer.RecordMetadata;
+import org.apache.kafka.common.PartitionInfo;
+import org.apache.thrift.TException;
+import org.apache.thrift.TSerializer;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.util.List;
+import java.util.Map;
+import java.util.Random;
```
yuyang08: remove this line?  Random is not used in this file

singer-commons/src/main/java/com/pinterest/singer/loggingaudit/client/AuditEventKafkaSender.java
```diff
@@ -0,0 +1,340 @@
...
+   *  umOfPartitionsToTrySending times (default 3 times) before dropping this events. Metrics are
+   *  used to track the queue size and usuage, number of events sent out to Kafka successfully, and
+   *  the number of events dropped.
+   */
+  @Override
+  public void run() {
+    LoggingAuditEvent event = null;
+    ProducerRecord<byte[], byte[]> record;
+    byte[] value = null;
+    List<PartitionInfo> partitionInfoList = this.kafkaProducer.partitionsFor(topic);
```
yuyang08: Could we refresh the partition info periodically to prevent us from using staled partition info?

singer-commons/src/main/java/com/pinterest/singer/loggingaudit/client/AuditEventKafkaSender.java
```diff
@@ -0,0 +1,340 @@
...
+  private Map<Integer, AtomicInteger> partitionErrorCounts = new ConcurrentHashMap<>();
+
+  /**
+   * For each event (identified by LoggingAuditHeaders, key of the map), track the partitions that
+   * has been used to send this events. These partitions are aka bad partitions, because it failed
+   * sending out this event. During retry, new partition not belong the bad partitions  is
+   * randomly chosen. When event is send out successfully or dropped, the corresponding entry in
+   * this map is removed.
+   */
+  private Map<LoggingAuditHeaders, Set<Integer>> eventBadPartitionsMap = new ConcurrentHashMap<>();
```
yuyang08: As we have only one audit topic, do we need to maintain a [ loggingauditheader --> set<bad partitions>] mapping?  if one partition gets offline,  all  auditing messages that send to that partition will fail.   shall we just have a `Set<Integer>` to track the bad partitions?

singer-commons/src/main/thrift/loggingaudit.thrift
```diff
@@ -38,3 +54,93 @@ struct LoggingAuditHeaders {
...
+
+   2: required LoggingAuditStage stage;
+
+   /**
+    *  flag indicates whether current stage is the start of logging auditing. For ThriftLogger stage,
+    *  the value is always true. At Singer stage, this field can be true if Singer uploads log files
+    *  not generated by ThriftLogger, for example certain server logs in text format. For other
+    *  stages, this field is always false.
+    */
+   3: required bool startAtCurrStage;
```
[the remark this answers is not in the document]
zzhhhzz: made the change in the latest update.

singer-commons/src/main/java/com/pinterest/singer/loggingaudit/client/AuditEventKafkaSender.java
zzhhhzz: removed in the latest update.

singer-commons/src/main/java/com/pinterest/singer/loggingaudit/client/AuditEventKafkaSender.java
zzhhhzz: implements refresh every 5 mins

singer-commons/src/main/java/com/pinterest/singer/loggingaudit/client/AuditEventKafkaSender.java
zzhhhzz: refactored in the latest update:

(1) maintained badPartitions: if event was not sent to one partition successfully, this partition will be added to badPartitions; if succeed, the partition is removed if it was added before.
badPartitions is for all events.  Every retry will choose a different partition that is not in badPartitions. 

(2) maintained  eventTriedCount: for each event, count number of tries, if reach the threshold 3, drop this event.  Note that: each retry is sent out to a partition that is randomly chosen and not in badPartitions;  also, event will be put into this map eventTriedCount only if it failed, this is done in the callback.

singer-commons/src/main/java/com/pinterest/singer/loggingaudit/client/AuditEventKafkaSender.java
```diff
@@ -0,0 +1,426 @@
...
+    LOG.warn(
+        "[{}] waits up to {} seconds to let [{}] send out LoggingAuditEvents left in the queue if"
+            + " any.",
+        Thread.currentThread().getName(), stopGracePeriodInSeconds, name);
+    int i = 0;
+    int numOfRounds = stopGracePeriodInSeconds / callingThreadSleepInSeconds;
+    while (queue.size() > 0 && this.thread != null && thread.isAlive() && i < numOfRounds) {
+      i += 1;
+      try {
+        Thread.sleep(callingThreadSleepInSeconds * 1000);
```
yuyang08: with this `Thread.sleep(....)`,  the process will wait for 5 minutes before it can exit. this can be problematic for some services.  shall we reduce the default waiting time to 10 seconds?   if there is message delivery failure within 10 seconds, in most of the cases having 300 seconds waiting time will not be helpful as the service will not be able to recovery within 5 minutes, given that we have already had retries on sending messages to different partitions.

singer-commons/src/main/java/com/pinterest/singer/loggingaudit/client/LoggingAuditClient.java
```diff
@@ -0,0 +1,258 @@
...
+                "stage=" + stage.toString());
+      }
+      if (successful) {
+        int qSize = queue.size();
+        double qUsagePercent = qSize * 1.0 / (qSize + queue.remainingCapacity());
+        OpenTsdbMetricConverter
+            .gauge(LoggingAuditClientMetrics.AUDIT_CLIENT_QUEUE_SIZE, qSize, "host=" + host,
+                "stage=" + stage.toString());
+        OpenTsdbMetricConverter
+            .gauge(LoggingAuditClientMetrics.AUDIT_CLIENT_QUEUE_USAGE_PERCENT, qUsagePercent,
```
yuyang08: if we have metrics on `AUDIT_CLIENT_QUEUE_USAGE_PERCENT`, do we still need `AUDIT_CLIENT_QUEUE_SIZE`?

singer-commons/src/main/java/com/pinterest/singer/loggingaudit/client/utils/KafkaUtils.java
```diff
@@ -0,0 +1,156 @@
...
+  public static KafkaProducer<byte[], byte[]> createKafkaProducer(KafkaProducerConfig config, String name) {
+    String brokerList = Joiner.on(',').join(config.getBrokerLists());
+    Properties properties = new Properties();
+    properties.put(
+        ProducerConfig.CLIENT_ID_CONFIG, name + "_" + UUID.randomUUID());
+    properties.put(ProducerConfig.BOOTSTRAP_SERVERS_CONFIG, brokerList);
+    properties.put(ProducerConfig.KEY_SERIALIZER_CLASS_CONFIG, config.getKeySerializerClass());
+    properties.put(ProducerConfig.VALUE_SERIALIZER_CLASS_CONFIG, config.getValueSerializerClass());
+
+    if (config.isTransactionEnabled()) {
```
yuyang08: a lot of code is duplicated with the code in singer/src/main/java/com/pinterest/singer/utils/KafkaUtils.java. can we avoid the duplication?

zzhhhzz: The Producer creation code is borrowed from the singer/src/main/java/com/pinterest/singer/utils/KafkaUtils.java.  To void de-duplicate, it requires moving packages/files in Singer module to Singer-commons module.  Given the current logging-audit branch is quite away from the master branch,  I feel it is safer to do the refactoring and avoiding duplicate AFTER  the logigng-audit branch is merged into the master branch.

singer-commons/src/main/java/com/pinterest/singer/loggingaudit/client/LoggingAuditClient.java
zzhhhzz: Queue capacity is a configuration property for LoggingAuditClient. For different use cases with different QPS,  queue capacity can be very different.   AUDIT_CLIENT_QUEUE_USAGE_PERCENT and  AUDIT_CLIENT_QUEUE_SIZE together can show the status of the queue and this will help to tune or set the queue_capacity property.   Therefore, I think both metrics are useful.

singer-commons/example/loggingaudit_client_config_thriftlogger_example.properties
```diff
@@ -0,0 +1,42 @@
...
+sender.kafka.producerConfig.ssl.keystore.password=your_password
+sender.kafka.producerConfig.ssl.keystore.type=JKS
+sender.kafka.producerConfig.ssl.secure.random.implementation=SHA1PRNG
+sender.kafka.producerConfig.ssl.truststore.location=/path/to/truststore/location
+sender.kafka.producerConfig.ssl.truststore.password=your_password
+sender.kafka.producerConfig.ssl.truststore.type=JKS
+
+auditedTopics.names=test1,test2,test3
+auditedTopics.test1.samplingRate=0.01
+auditedTopics.test1.startAtCurrentStage=true
```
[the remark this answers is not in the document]
zzhhhzz: Both options, stopAtCurrentStage (startAtCurrentStage) and stop_stage (start_stage),  were considered initially.  Choosing stopAtCurrentStage over stop_stage is because there can be different systems (in addition to ThriftLogger, Singer Merced) importing this library for auditing and the number of use cases can increase,  thus we don't know what will be added or removed in the future.  For example, certain streaming app (kafka streams, flink, spark stream)  interacting with Kafka can join the auditing as well.  Thus, I feel specifying stopAtCurrentStage rather stop_stage can provide more flexibility and extensibility.

singer-commons/src/main/java/com/pinterest/singer/loggingaudit/client/AuditEventKafkaSender.java
```diff
@@ -0,0 +1,426 @@
...
+    return kafkaProducer;
+  }
+
+  public void setKafkaProducer(KafkaProducer<byte[], byte[]> kafkaProducer) {
+    this.kafkaProducer = kafkaProducer;
+  }
+
+  public int getAlternatePartition(int numOfPartitions) {
+    int randomPartition = 0;
+    int trials = 10;
```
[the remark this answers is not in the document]
zzhhhzz: made the change in the latest update.

singer-commons/src/main/java/com/pinterest/singer/loggingaudit/client/AuditEventKafkaSender.java
```diff
@@ -0,0 +1,426 @@
...
+      if (!badPartitions.contains(randomPartition)) {
+        break;
+      }
+    }
+    return randomPartition;
+  }
+
+  private void refreshPartitionIfNeeded() {
+    // refresh every 5 mins
+    if (System.currentTimeMillis() - lastTimeUpdate > 1000 * 300) {
```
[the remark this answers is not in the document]
zzhhhzz: made the change in the latest update.

singer-commons/example/loggingaudit_client_config_thriftlogger_example.properties
```diff
@@ -0,0 +1,42 @@
...
+
+
+stage=thriftlogger
+enableAuditForAllTopicsByDefault=false
+queueSize=100000
+enqueueWaitInMilliseconds=0
+
+sender.type=kafka
+sender.kafka.topic=logging_audit
+sender.kafka.producerConfig.bootstrap.servers=testkafka-001:9092,testkafka-002:9092,testkafka-003:9092
```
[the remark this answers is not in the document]
zzhhhzz: Using producerConfig is consistent with current naming used in Singer module.  If we want to shorten the name,  this could be done after the logging-audit branch is merged. Refactoring, avoiding duplicate code and simplifying naming can be done easier.

singer-commons/src/main/java/com/pinterest/singer/loggingaudit/client/AuditEventKafkaSender.java
```diff
@@ -0,0 +1,426 @@
...
+   *  umOfPartitionsToTrySending times (default 3 times) before dropping this events. Metrics are
+   *  used to track the queue size and usuage, number of events sent out to Kafka successfully, and
+   *  the number of events dropped.
+   */
+  @Override
+  public void run() {
+    LoggingAuditEvent event = null;
+    ProducerRecord<byte[], byte[]> record;
+    byte[] value = null;
+    refreshPartitionIfNeeded();
```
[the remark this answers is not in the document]
zzhhhzz: Note that.  In refreshPartitionIfNeeded(), there is try...catch block and failing to refresh the List<PartitionInfo> will not break/exit the application; metrics are used to track the refresh success / failures.  This is the expected behavior in the while loop (LINE 227) because we can still reply on the old (not refreshed) List<PartitionInfo> to process more events.

In line 221, this call is to initialize List<PartitionInfo> and a Preconditions checking is right after this line to make sure, the list is not empty in order to entry the WHILE loop.  I feel it is better to keep it.

singer-commons/src/main/java/com/pinterest/singer/loggingaudit/client/AuditEventKafkaSender.java
zzhhhzz: 5 mins is the maximum amount of time to wait.  If there is no events in the queue or a small number of events that quickly get delivered in a short period (say 90 seconds), the wait will stop earlier.  Reserving some time during the shutdown is just trying the best to avoid audit events loss in order to give more accurate auditing results later. 

Since different systems (services using thrift logger, Singer, Merced, etc) can tolerate different amount of time for grace shutdown,  I will make this a configuration property with default 5 mins.  If certain systems don't want that long, a smaller value should be specified.

singer-commons/example/loggingaudit_client_config_thriftlogger_example.properties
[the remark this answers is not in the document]
yuyang08: Thanks for the explanation! Could you include this  in the comment?  that will help later readers to understand the design thoughts.

singer-commons/src/main/java/com/pinterest/singer/loggingaudit/client/AuditEventKafkaSender.java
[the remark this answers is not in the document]
yuyang08: what is the value of having `Preconditions.checkArgument(partitionInfoList.size() > 0)` checking?   if the condition fails, will the program throws exception and terminate?   in general, we should only keep the code that is really needed. otherwise it will cause confusion to other readers later.

singer-commons/src/main/java/com/pinterest/singer/loggingaudit/client/utils/KafkaUtils.java
yuyang08: We should do the refactoring upfront, instead of after logging-audit branch is merged into the master branch.

zzhhhzz: refactored KafkaUtils.java in the update of the PR to maintain only one piece of code.

singer-commons/src/main/java/com/pinterest/singer/loggingaudit/client/AuditEventKafkaSender.java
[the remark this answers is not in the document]
zzhhhzz: removed line 221 to line 223.  The refresh method in the WHILE loop is enough. Made the change in the latest update of this PR.

singer-commons/pom.xml
zzhhhzz: changed together with the Kafka dependency

thrift-logger/src/main/java/com/pinterest/singer/client/logback/AuditableLogbackThriftLoggerFactory.java
ambud: Still recommend we should use instance variables and static reference in SingerSettings etc.

singer-commons/src/main/java/com/pinterest/singer/loggingaudit/client/LoggingAuditClient.java
```diff
@@ -0,0 +1,258 @@
...
+      LOG.info("[{}] add TopicAuditConfig ({}) for {}.", Thread.currentThread().getName(),
+          auditConfig.toString(), loggingAuditName);
+    }
+    return auditConfig;
+  }
+
+  /**
+   *  check if there exists audit config for a topic / loggingAuditName
+   */
+  public boolean auditConfigExists(String loggingAuditName) {
```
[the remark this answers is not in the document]
zzhhhzz: made the change in the update of the PR.

singer-commons/src/main/java/com/pinterest/singer/loggingaudit/client/AuditEventKafkaSender.java
```diff
@@ -0,0 +1,416 @@
...
+  /**
+   *  If sending out to one partition fails, this partition is added to set;
+   *  If sending out to one partition succeeds, this partition is removed if it was added before.
+   */
+  private Set<Integer> badPartitions = ConcurrentHashMap.newKeySet();
+
+  /**
+   * For each event (identified by LoggingAuditHeaders, key of the map), track the number of tries
+   * for sending to Kafka. Each try will choose a different and partition that is not in the
+   * badPartitions set.  When event is send out successfully or dropped, the corresponding entry in
```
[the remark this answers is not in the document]
zzhhhzz: Here I mean badPartitions hashset.  I will remove set to eliminate the confusion here

singer-commons/src/main/java/com/pinterest/singer/loggingaudit/client/AuditEventKafkaSender.java
```diff
@@ -0,0 +1,416 @@
...
+            .incr(LoggingAuditClientMetrics.AUDIT_CLIENT_SENDER_KAFKA_PARTITIONS_REFRESH_ERROR, 1,
+                "host=" + host, "stage=" + stage.toString());
+      }
+    }
+  }
+
+  /**
+   *  Sender dequeue LoggingAuditEvents and send out to Kafka cluster. If send out to one partition
+   *  fails, it will choose another partition. For each event, it will try at most
+   *  umOfPartitionsToTrySending times (default 3 times) before dropping this events. Metrics are
```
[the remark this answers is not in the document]
zzhhhzz: made the change in the update of the PR

singer-commons/src/main/java/com/pinterest/singer/loggingaudit/client/AuditEventKafkaSender.java
```diff
@@ -0,0 +1,416 @@
...
+      } catch (Exception e) {
+        OpenTsdbMetricConverter
+            .incr(LoggingAuditClientMetrics.AUDIT_CLIENT_SENDER_KAFKA_PARTITIONS_REFRESH_ERROR, 1,
+                "host=" + host, "stage=" + stage.toString());
+      }
+    }
+  }
+
+  /**
+   *  Sender dequeue LoggingAuditEvents and send out to Kafka cluster. If send out to one partition
```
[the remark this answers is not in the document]
zzhhhzz: make the change in the update of the PR

singer-commons/src/main/java/com/pinterest/singer/loggingaudit/client/AuditEventKafkaSender.java
```diff
@@ -0,0 +1,416 @@
...
+    public void onCompletion(RecordMetadata recordMetadata, Exception e) {
+      try {
+        int partition = recordMetadata.partition();
+        if (e == null) {
+          OpenTsdbMetricConverter
+              .incr(LoggingAuditClientMetrics.AUDIT_CLIENT_SENDER_KAFKA_EVENTS_ACKED, 1,
+                  "host=" + host, "stage=" + stage.toString(),
+                  "logName=" + event.getLoggingAuditHeaders().getLogName());
+
+          // if send is successfully, remove the event from the map eventTriedCount if it was added
```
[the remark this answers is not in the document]
zzhhhzz: make the change in the update of the PR

singer-commons/src/main/java/com/pinterest/singer/loggingaudit/client/AuditEventKafkaSender.java
```diff
@@ -0,0 +1,416 @@
...
+  public class KafkaProducerCallback implements Callback {
+
+    private LoggingAuditEvent event;
+
+    public KafkaProducerCallback(LoggingAuditEvent event) {
+      this.event = event;
+    }
+
+    @Override
+    public void onCompletion(RecordMetadata recordMetadata, Exception e) {
```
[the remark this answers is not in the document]
zzhhhzz: since inside this method, when exception is not null (send failed), we are doing a lot of checks. This part can be factored out as a separate method. I can make this change in the update of the PR

singer/src/main/java/com/pinterest/singer/writer/KafkaWriter.java
```diff
@@ -271,7 +288,8 @@ private void includeAuditMessageInBatch(List<ProducerRecord<byte[], byte[]>> mes
 
       ProducerRecord<byte[], byte[]> keyedMessage;
       for (int i = 0; i < validPartitions.size(); i++) {
-        buckets.add(new ArrayList<>());
+        buckets.put(validPartitions.get(i).partition(), new ArrayList<>());
```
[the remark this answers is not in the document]
zzhhhzz: explicitly create a index variable and document it.  make this change in the update of the PR

singer/src/main/java/com/pinterest/singer/writer/KafkaWriter.java
```diff
@@ -338,7 +365,23 @@ public void writeLogMessages(List<LogMessage> logMessages) throws LogStreamWrite
           // get the max write latency
           maxKafkaBatchWriteLatency = Math.max(maxKafkaBatchWriteLatency,
               result.getKafkaBatchWriteLatencyInMillis());
+          if (this.enableLoggingAudit && SingerSettings.getLoggingAuditClient() != null) {
```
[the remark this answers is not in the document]
zzhhhzz: create a separate method isLoggingAuditEnabledAndConfigured() to make this part clean. make the change in the update of the PR

singer-commons/src/main/java/com/pinterest/singer/loggingaudit/client/AuditEventKafkaSender.java
```diff
@@ -0,0 +1,416 @@
...
+        badPartitions.add(partition);
+        OpenTsdbMetricConverter
+            .incr(LoggingAuditClientMetrics.AUDIT_CLIENT_SENDER_KAFKA_PARTITION_ERROR, 1,
+                "host=" + host, "stage=" + stage.toString(), "topic=" + topic,
+                "partition=" + partition);
+
+        // retry the failed event by inserting it at the beginning of the deque.
+        // If number of tries reaches 3, meaning that 3 partitions have been tried sending to but
+        // still failed, this event is dropped.
+        if (!eventTriedCount.containsKey(event.getLoggingAuditHeaders())) {
```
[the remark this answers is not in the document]
zzhhhzz: change to get() first and do the if...else based on whether returned value is null or not.  Make this change in the update of the PR

thrift-logger/src/main/java/com/pinterest/singer/client/logback/AuditableLogbackThriftLoggerFactory.java
zzhhhzz: I feel it is easier to make it static in AuditableLogbackThriftLoggerFactory.  The actual logger can easily reference the audit client when trying to send the audit events.

singer-commons/src/main/java/com/pinterest/singer/loggingaudit/client/AuditEventKafkaSender.java
```diff
@@ -0,0 +1,420 @@
...
+   *   stopGracePeriodInSeconds / THREAD_SLEEP_IN_SECONDS is the total rounds the calling thread
+   *   will sleep.
+   */
+  private static final int THREAD_SLEEP_IN_SECONDS = 10;
+
+  /**
+   *  when gracefully shutting down the Sender, this field specifies maximum time for main thread
+   *  to wait, in order to let the sender send out audit events left in the queue if there are any.
+   */
+  private int stopGracePeriodInSeconds = 300;
```
yuyang08: can we lower the default value to 15 seconds?    300 seconds is too long. we don't want to be in the blocking path during java service restart.

singer-commons/src/main/thrift/loggingaudit_config.thrift
```diff
@@ -0,0 +1,122 @@
...
+}
+
+
+/**
+ *  KakfaSenderConfig consists of kafka topic name and kafka producer config.
+ **/
+struct KafkaSenderConfig {
+    1: optional string topic = "logging_audit";
+    2: optional common.KafkaProducerConfig kafkaProducerConfig;
+    3: optional i32 stopGracePeriodInSeconds = 300;
```
yuyang08: change the default value to 30 seconds?  300 seconds can be too long  for java service restart.

singer-commons/src/main/java/com/pinterest/singer/loggingaudit/client/AuditEventKafkaSender.java
zzhhhzz: it makes sense.  while the total grace period is configurable, making the default to be 30 seconds is good in terms of not blocking service restart.  I made the change in the update of the PR

singer-commons/src/main/thrift/loggingaudit_config.thrift
zzhhhzz: I made the change in the update of the PR

singer/src/main/java/com/pinterest/singer/writer/KafkaWriter.java
```diff
@@ -221,6 +233,11 @@ public HeadersInjector getHeadersInjector() {
     return headersInjector;
   }
 
+  @VisibleForTesting
+  public ExecutorService getClusterThreadPool() {
+    return clusterThreadPool;
+  }
+
```
yuyang08: Could we remove the method `includeAuditMessageInBatch` as it is no longer used.

singer/src/main/java/com/pinterest/singer/writer/KafkaWriter.java
```diff
@@ -271,7 +288,11 @@ private void includeAuditMessageInBatch(List<ProducerRecord<byte[], byte[]>> mes
 
       ProducerRecord<byte[], byte[]> keyedMessage;
       for (int i = 0; i < validPartitions.size(); i++) {
-        buckets.add(new ArrayList<>());
+        // for each partitionId, there is a corresponding bucket in buckets and a corresponding
+        // headersMap in mapOfHeadersMaps.
+        int partitionId = validPartitions.get(i).partition();
+        buckets.put(partitionId, new ArrayList<>());
+        mapOfHeadersMaps.put(partitionId ,new HashMap<Integer, LoggingAuditHeaders>());
```
yuyang08: nit: simplify as `        mapOfHeadersMaps.put(partitionId ,new HashMap<>());`  as `Integer, LoggingAuditHeaders` part is really not needed

singer/src/main/java/com/pinterest/singer/common/SingerSettings.java
```diff
@@ -196,6 +207,18 @@ public static void initialize(SingerConfig config)
     globalFsm.start();
   }
 
+  protected static void loadAndSetLoggingAuditClient(SingerConfig config) {
```
yuyang08: shall we name the method `loadAndSetLoggingAuditClientIfEnabled`?

singer/src/main/java/com/pinterest/singer/writer/KafkaWriter.java
```diff
@@ -328,18 +364,39 @@ public void writeLogMessages(List<LogMessage> logMessages) throws LogStreamWrite
...
-          throw new LogStreamWriterException("Failed to write messages to kafka");
+          anyBucketSendFailed = true;
         } else {
           bytesWritten += result.getWrittenBytesSize();
           // get the max write latency
           maxKafkaBatchWriteLatency = Math.max(maxKafkaBatchWriteLatency,
               result.getKafkaBatchWriteLatencyInMillis());
+          if (isLoggingAuditEnabledAndConfigured()) {
+            int bucketIndex = result.getPartition();
+            if (bucketIndex >= 0 && result.getRecordMetadataList().size() != buckets.get(bucketIndex).size()){
```
yuyang08: could you add comment to explain why we need this check ?

singer/src/main/java/com/pinterest/singer/writer/KafkaWriter.java
zzhhhzz: removed in the update of the PR.

singer/src/main/java/com/pinterest/singer/writer/KafkaWriter.java
zzhhhzz: added comments in the update of the PR

singer/src/main/java/com/pinterest/singer/common/SingerSettings.java
zzhhhzz: made the change in the update of the PR

singer/src/main/java/com/pinterest/singer/writer/KafkaWriter.java
zzhhhzz: made the change in the update of the PR.  let java compiler to infer the type at the runtime

thrift-logger/src/main/java/com/pinterest/singer/client/AuditableLogbackThriftLogger.java
```diff
@@ -132,23 +160,35 @@ private void appendInternal(byte[] partitionKey, byte[] message, long timeNanos,
     try {
```
yuyang08: How about have the following change to minimize boilerplate code?

diff --git a/thrift-logger/src/main/java/com/pinterest/singer/client/AuditableLogbackThriftLogger.java b/thrift-logger/src/main/java/com/pinterest/singer/client/AuditableLogbackThriftLogger.java
index 5a8b0dc..13a9302 100644
--- a/thrift-logger/src/main/java/com/pinterest/singer/client/AuditableLogbackThriftLogger.java
+++ b/thrift-logger/src/main/java/com/pinterest/singer/client/AuditableLogbackThriftLogger.java
@@ -131,7 +131,7 @@ public class AuditableLogbackThriftLogger extends LogbackThriftLogger {
     if (enableLoggingAudit) {
       headers = auditHeadersGenerator.generateHeaders();
     }
-    appendInternal(partitionKey, message,timeNanos, headers);
+    append(partitionKey, message,timeNanos, headers);
   }
 
   @Override
@@ -147,7 +147,7 @@ public class AuditableLogbackThriftLogger extends LogbackThriftLogger {
         }
       }
       byte[] messageBytes = ThriftCodec.getInstance().serialize(thriftMessage);
-      appendInternal(partitionKey, messageBytes, timeNanos, headers);
+      append(partitionKey, messageBytes, timeNanos, headers);
     } catch (TException e) {
       OpenTsdbMetricConverter.incr(THRIFT_LOGGER_ERROR_TEXCEPTION,
           "topic=" + topic, "host=" + HOST_NAME);
@@ -155,36 +155,23 @@ public class AuditableLogbackThriftLogger extends LogbackThriftLogger {
     }
   }
 
-  private void appendInternal(byte[] partitionKey, byte[] message, long timeNanos,
+  private void append(byte[] partitionKey, byte[] message, long timeNanos,
                               LoggingAuditHeaders headers) throws LogbackException {
-    try {
       LogMessage logMessage = new LogMessage()
-          .setTimestampInNanos(timeNanos)
-          .setMessage(message);
+          .setTimestampInNanos(timeNanos).setMessage(message);
 
       if(this.enableLoggingAudit && headers != null) {
         logMessage.setLoggingAuditHeaders(headers);
         OpenTsdbMetricConverter.incr(AUDIT_THRIFT_LOGGER_HEADERS_ADDED_TO_LOG_MESSAGE_COUNT,
             "topic=" + topic, "host=" + HOST_NAME);
       }
-
-      if (partitionKey != null) {
-        logMessage.setKey(partitionKey);
-      }
-      appender.doAppend(logMessage);
-      OpenTsdbMetricConverter.incr(THRIFT_LOGGER_COUNT_METRIC, "topic=" + topic, "host=" + HOST_NAME);
-
+      super.append(logMessage, partitionKey);
       if (this.enableLoggingAudit && headers != null && AuditableLogbackThriftLoggerFactory.getLoggingAuditClient() != null) {
           AuditableLogbackThriftLoggerFactory.getLoggingAuditClient().audit(this.topic, headers, true,
               System.currentTimeMillis());
           OpenTsdbMetricConverter.incr(AUDIT_THRIFT_LOGGER_AUDITED_MESSAGE_COUNT,
               "topic=" + topic, "host=" + HOST_NAME);
-
       }
-    } catch (LogbackException e) {
-      OpenTsdbMetricConverter.incr(THRIFT_LOGGER_ERROR_LOGBACKEXCEPTION, "topic=" + topic, "host=" + HOST_NAME);
-      throw e;
-    }
   }
 
   public boolean shouldAudit(){
diff --git a/thrift-logger/src/main/java/com/pinterest/singer/client/LogbackThriftLogger.java b/thrift-logger/src/main/java/com/pinterest/singer/client/LogbackThriftLogger.java
index 1934f4f..5c5fa98 100644
--- a/thrift-logger/src/main/java/com/pinterest/singer/client/LogbackThriftLogger.java
+++ b/thrift-logger/src/main/java/com/pinterest/singer/client/LogbackThriftLogger.java
@@ -70,13 +70,13 @@ public class LogbackThriftLogger extends BaseThriftLogger {
     this.appender = appender;
   }
 
-  @Override
-  public void append(byte[] partitionKey, byte[] message, long timeNanos) {
+  /**
+   * This method is also called in AuditableLogackThriftLogger to minimize boilerplate code.
+   * @param logMessage  logMessage
+   * @param partitionKey partition key
+   */
+  protected void append(LogMessage logMessage, byte[] partitionKey) {
     try {
-      LogMessage logMessage = new LogMessage()
-          .setTimestampInNanos(timeNanos)
-          .setMessage(message);
-
       if (partitionKey != null) {
         logMessage.setKey(partitionKey);
       }
@@ -90,6 +90,12 @@ public class LogbackThriftLogger extends BaseThriftLogger {
     }
   }
 
+  @Override
+  public void append(byte[] partitionKey, byte[] message, long timeNanos) {
+      LogMessage logMessage = new LogMessage().setTimestampInNanos(timeNanos).setMessage(message);
+      append(logMessage, partitionKey);
+  }
+
   @Override
   public void append(byte[] partitionKey, TBase thriftMessage, long timeNanos) throws TException {
     try {

thrift-logger/src/main/java/com/pinterest/singer/client/logback/AuditableLogbackThriftLoggerFactory.java
```diff
@@ -36,32 +51,149 @@
```
yuyang08: shall we have`AuditableLogbackThriftLoggerFactory` extends `LogbackThriftLoggerFactory`?

zzhhhzz: refactored in the update of the PR.

thrift-logger/src/main/java/com/pinterest/singer/client/AuditableLogbackThriftLogger.java
zzhhhzz: I applied your diff and removed boilerplate code. thanks for this great suggestion.

singer-commons/src/main/java/com/pinterest/singer/loggingaudit/client/AuditEventKafkaSender.java
```diff
@@ -0,0 +1,420 @@
...
+      badPartitions.add(recordMetadata.partition());
+      OpenTsdbMetricConverter
+          .incr(LoggingAuditClientMetrics.AUDIT_CLIENT_SENDER_KAFKA_PARTITION_ERROR, 1,
+              "host=" + host, "stage=" + stage.toString(), "topic=" + topic,
+              "partition=" + recordMetadata.partition());
+
+      // retry the failed event by inserting it at the beginning of the deque.
+      // If number of tries reaches 3, meaning that 3 partitions have been tried sending to but
+      // still failed, this event is dropped.
+
```
yuyang08: remove the empty line here?

singer-commons/src/main/java/com/pinterest/singer/loggingaudit/client/AuditEventKafkaSender.java
```diff
@@ -0,0 +1,420 @@
...
+      Integer count = eventTriedCount.get(event.getLoggingAuditHeaders());
+      if (count == null){
+        eventTriedCount.put(event.getLoggingAuditHeaders(), 1);
+        OpenTsdbMetricConverter
+            .gauge(LoggingAuditClientMetrics.AUDIT_CLIENT_SENDER_KAFKA_EVENTS_RETRIED,
+                eventTriedCount.size(), "host=" + host, "stage=" + stage.toString(),
+                "topic=" + topic);
+      } else {
```
yuyang08: compline the `else { if ..` as  one `else if (..` ?

singer-commons/src/main/java/com/pinterest/singer/loggingaudit/client/AuditEventKafkaSender.java
```diff
@@ -0,0 +1,420 @@
...
+          LOG.debug("Failed to send audit event after trying {} partitions. Drop this event.",
+              count);
+          OpenTsdbMetricConverter
+              .incr(LoggingAuditClientMetrics.AUDIT_CLIENT_SENDER_KAFKA_EVENTS_DROPPED, 1,
+                  "host=" + host, "stage=" + stage.toString(),
+                  "logName=" + event.getLoggingAuditHeaders().getLogName());
+          eventTriedCount.remove(event.getLoggingAuditHeaders());
+        } else {
+          eventTriedCount.put(event.getLoggingAuditHeaders(), count + 1);
+          boolean success = false;
```
yuyang08: remove line 306 and update line 308 as 

    boolean success = queue.offerFirst(event, 3, TimeUnit.SECONDS);

singer/src/main/java/com/pinterest/singer/writer/KafkaWritingTask.java
```diff
@@ -51,18 +53,25 @@ public KafkaWritingTask(KafkaProducer<byte[], byte[]> producer,
     this.taskCreationTimeInMillis = System.currentTimeMillis();
     try {
       leaderNode = sortedPartitions.get(msgs.get(0).partition()).leader().host();
+      partition = sortedPartitions.get(msgs.get(0).partition()).partition();
```
yuyang08: update the two lines above to reduce redundant function calls?

        PartitionInfo firstPartition = sortedPartitions.get(msgs.get(0).partition());
        leaderNode = firstPartition.leader().host();
        partition = firstPartition.partition();

singer-commons/src/main/java/com/pinterest/singer/loggingaudit/client/AuditEventKafkaSender.java
zzhhhzz: removed in the update of the PR.

singer-commons/src/main/java/com/pinterest/singer/loggingaudit/client/AuditEventKafkaSender.java
zzhhhzz: made the change in the update of the PR

singer-commons/src/main/java/com/pinterest/singer/loggingaudit/client/AuditEventKafkaSender.java
zzhhhzz: made the change in the update of the PR.

singer/src/main/java/com/pinterest/singer/writer/KafkaWritingTask.java
zzhhhzz: made the change in the update of the PR.

PR #17, opened 2019-11-08
Add support injection of Pod Uid instead of hostname for text logger

ambud:
Add support injection of Pod Uid instead of hostname for text logger. Logic:

If prepend hostname is enabled for text logger config & singer is running in Kubernetes mode then attempt to use pod id as hostname if pod if is NOT null or NON_KUBERNETES_POD_ID else return hostname.

```Results :

Tests run: 65, Failures: 0, Errors: 0, Skipped: 0

[INFO] ------------------------------------------------------------------------
[INFO] Reactor Summary for Singer Logging Agent 0.7.3.29:
[INFO] 
[INFO] Singer Logging Agent ............................... SUCCESS [  0.138 s]
[INFO] thrift-logger ...................................... SUCCESS [  8.141 s]
[INFO] singer ............................................. SUCCESS [02:32 min]
[INFO] ------------------------------------------------------------------------
[INFO] BUILD SUCCESS
[INFO] ------------------------------------------------------------------------
[INFO] Total time:  02:40 min
[INFO] Finished at: 2019-11-08T18:44:21Z
[INFO] ------------------------------------------------------------------------```

### DIFF
--- a/singer/src/main/java/com/pinterest/singer/utils/SingerUtils.java
+++ b/singer/src/main/java/com/pinterest/singer/utils/SingerUtils.java
@@ -15,6 +15,7 @@
  */
 package com.pinterest.singer.utils;
 
+import com.pinterest.singer.common.LogStream;
 import com.pinterest.singer.common.SingerSettings;
 import com.pinterest.singer.config.DirectorySingerConfigurator;
 import com.pinterest.singer.config.PropertyFileSingerConfigurator;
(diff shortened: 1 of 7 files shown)

### CODE REVIEW

File: singer/src/main/java/com/pinterest/singer/utils/SingerUtils.java [hunk 1]
```diff
@@ -300,4 +298,16 @@ public static String makeGetRequest(String uri) throws IOException {
           throw new IOException(e);
       }
   }
+
+  public static String getHostNameBasedOnConfig(LogStream logStream,
+                                                     SingerConfig singerConfig) {
+    if (singerConfig.isKubernetesEnabled()) {
+      if (logStream.getSingerLog().getPodUid() !=null
```
yuyang08: when will we get `logStream.getSingerLog().getPodUid() == null`?

ambud: LogStreamManager:L252 since Singer Kube support both local and pods this corner case needs to be covered.

PR #21, opened 2019-11-26
Null check in KafkaWriterTask

kabochya:
Currently if the call fails before result is populated, the finally block will fail due to NPE. Add a sanity check to avoid this.


singer/src/main/java/com/pinterest/singer/writer/KafkaWritingTask.java
```diff
@@ -116,7 +116,7 @@ public KafkaWritingTaskResult call() {
       OpenTsdbMetricConverter.incrGranular(SingerMetrics.BROKER_WRITE_FAILURE, 1, "broker=" + leaderNode);
       result = new KafkaWritingTaskResult(false, e);
     } finally {
-      if (!result.success) {
+      if (result != null && !result.success) {
```
yuyang08: why do we need to add `result != null` check here?

kabochya: A Throwable happening before L93 will not be catched. `result` will not be initialized and causes result.success to throw an NPE that masks the actual Throwable from getting escalated.

ambud:
Closing this PR to be recreated.

PR #22, opened 2019-11-26
[WIP] Mapped Thrift Reader implementation

ambud:
Mapped Thrift Reader uses memory mapped files instead of RandomAccessFile to read thrift logs. The advantage of this design is the reduced buffer copy and object overhead for Singer thereby reducing net heap usage. Since the data is kept in page cache in both implementations this implementation doesn't increase any overheads.

### DIFF
--- a/singer/src/main/java/com/pinterest/singer/reader/thrift/MappedFileTBinaryProtocol.java
+++ b/singer/src/main/java/com/pinterest/singer/reader/thrift/MappedFileTBinaryProtocol.java
@@ -1,0 +1,407 @@
+/**
+ * Copyright 2019 Pinterest, Inc.
+ * 
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ * 
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ * 
+ * Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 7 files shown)


singer/src/main/java/com/pinterest/singer/reader/thrift/MappedFileTBinaryProtocol.java
```diff
@@ -0,0 +1,405 @@
...
+ * {@link FileChannel}. This only creates a memory mapping however data will
+ * only be read when actual bytes are accessed so this is safe from a memory
+ * utilization. It's the responsibility of the underlying operating system to
+ * swap pages if sufficient memory is not available.
+ * 
+ * Note that {@link TTransport} can't be used in this case since the
+ * {@link TProtocol} needs direct reference to the ByteBuffer for lower level
+ * byte manipulations which won't be possible without direct reference.
+ */
+public class MappedFileTBinaryProtocol extends TProtocol {
```
yuyang08: shall we extend from`TBinaryProtocol`?

ambud: That wouldn't add value since all methods have to be overridden. It calls TTransport methods for byte reads under the hood which we are trying to avoid since it defeats the purpose of this change. This change does special reads for ByteBuffer / readBinary by performing a slice creation.

ambud:
Code has been merged in separate PR.

PR #23, opened 2019-11-27
Add metric to track number of kafka producers, this will help us

ambud:
Add metric to track number of kafka producers, this will help us determine if we need to tune the producer buffers to lower the memory footprint. Each producer uses 32MB buffer.memory by default which can add up if Singer is writing to a lot of Kafka clusters. 

Note that this buffer is allocated to the heap.

### DIFF
--- a/singer/src/main/java/com/pinterest/singer/writer/KafkaProducerManager.java
+++ b/singer/src/main/java/com/pinterest/singer/writer/KafkaProducerManager.java
@@ -15,6 +15,8 @@
  */
 package com.pinterest.singer.writer;
 
+import com.pinterest.singer.common.SingerMetrics;
+import com.pinterest.singer.metrics.OpenTsdbMetricConverter;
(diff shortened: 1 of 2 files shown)


singer/src/main/java/com/pinterest/singer/writer/KafkaProducerManager.java
```diff
@@ -83,6 +85,8 @@ public static void shutdown() {
       }
     }
     result = producers.get(config);
+    // log metrics for no.of kafka producers currently in the cache
+    OpenTsdbMetricConverter.addMetric(SingerMetrics.NUM_KAFKA_PRODUCERS, producers.size());
```
HenryCaiHaiying: Why the metric tracking is done in the getter path not done in line 81 when the new producer is created?

ambud: Based on discussion added this metric to both create and reset producer

PR #33, opened 2020-01-07
Adding producer metric monitoring framework

ambud:
This allows us to monitor specific kafka producer metrics. This will help monitor performance, resource consumption and improve tuning.

### DIFF
--- a/singer/src/main/java/com/pinterest/singer/SingerMain.java
+++ b/singer/src/main/java/com/pinterest/singer/SingerMain.java
@@ -54,6 +54,14 @@ public final class SingerMain {
       } catch (Throwable t) {
         LOG.error("Shutdown failure: log monitor : ", t);
       }
+      
+      try {
+        if (SingerSettings.getKafkaProducerMonitorThread() != null) {
+          SingerSettings.getKafkaProducerMonitorThread().interrupt();
+        }
+      }catch(Throwable t) {
+        LOG.error("Shutdown error: kafka producer metrics monitor : ", t);
+      }
 
       try {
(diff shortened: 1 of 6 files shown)


singer/src/main/java/com/pinterest/singer/SingerMain.java
```diff
@@ -54,6 +54,14 @@ public void run() {
       } catch (Throwable t) {
         LOG.error("Shutdown failure: log monitor : ", t);
       }
+
```
zzhhhzz: Instead of introducing another daemon thread,  how about using ostrich to collect kafka producer metrics like all other metrics that are being collected now?

ambud: There are two downsides to that approach.

1. Metrics are conflicting and duplicate as a single producer may be shared by multiple kafkawriter instances
2. Lack of centralized management of producer level metrics

An independent thread makes for a clean and non-intrusive design.

ambud:
> Is another thread necessary or we can just add metric reporting code when Kafka producers are used?

Yes

singer/src/main/java/com/pinterest/singer/common/SingerSettings.java
```diff
@@ -148,6 +151,10 @@ public static void initialize(SingerConfig config)
           logWritingExecutors.put(clusterSig, threadPool);
         }
       }
+      
+      kafkaProducerMonitorThread = new Thread(new KafkaProducerMetricsMonitor());
```
vahidhashemian: Would we need to check if this static member is already initialized?

PR #38, opened 2020-02-12
Create monitor for Kafka Producer metrics and expose them as singer metrics

ambud:
Create monitor for Kafka Producer metrics and expose them as singer metrics. This allows us to view utilization and tune.


singer/pom.xml
```diff
@@ -189,6 +189,12 @@
             <artifactId>aws-java-sdk-s3</artifactId>
             <version>1.11.35</version>
         </dependency>
+        <dependency>
+		    <groupId>com.salesforce.kafka.test</groupId>
```
vahidhashemian: Is this a reliable dependency? The work `test` in the group id makes me wonder.

singer/src/main/java/com/pinterest/singer/writer/KafkaProducerMetricsMonitor.java
```diff
@@ -0,0 +1,86 @@
...
+          OpenTsdbMetricConverter.gauge(entry.getKey().name(), entry.getValue().value(),
+              "cluster=" + signature);
+        }
+      }
+    }
+  }
+
+  public static String convertSignatureToTag(KafkaProducerConfig key) {
+    return key.getKafkaClusterSignature().replace("/discovery/", "").replace("/prod", "")
+        .replace("/", "_");
```
vahidhashemian: Do we need to account for `tls` serversets in this conversion?

singer/pom.xml
ambud: Yes, it's called test because it's a package used for testing.

singer/src/main/java/com/pinterest/singer/writer/KafkaProducerMetricsMonitor.java
ambud: My current thought is to have other names be appended to the setup since the key is different so if it's any extension it really would represent a different cluster signature meaning a different producer.

singer/src/main/java/com/pinterest/singer/writer/KafkaProducerMetricsMonitor.java
```diff
@@ -0,0 +1,86 @@
...
+    Map<KafkaProducerConfig, KafkaProducer<byte[], byte[]>> producers = KafkaProducerManager
+        .getInstance().getProducers();
+    for (Entry<KafkaProducerConfig, KafkaProducer<byte[], byte[]>> kafkaProducerEntry : producers
+        .entrySet()) {
+      KafkaProducerConfig key = kafkaProducerEntry.getKey();
+      String signature = convertSignatureToTag(key);
+      Map<MetricName, ? extends Metric> metrics = kafkaProducerEntry.getValue().metrics();
+      for (Entry<MetricName, ? extends Metric> entry : metrics.entrySet()) {
+        if (PRODUCER_METRICS_WHITELIST.contains(entry.getKey().name())) {
+          OpenTsdbMetricConverter.gauge(entry.getKey().name(), entry.getValue().value(),
```
kabochya: It seems that the value() method is deprecated in 1.0.0 and metricValue() is the one getting used instead. Is there any reason of using value() over metricValue()?

ambud: Yes, metricValue returns an Object type which we will need to switch on and detect type before down casting.

PR #39, opened 2020-02-15
Make Kafka buffer.memory configurable

ambud:
Make Kafka buffer.memory configurable as this will allow us to reduce unnecessary buffer memory used in singer making it less memory hungry.

```
buffer.memory: The total bytes of memory the producer can use to buffer records waiting to be sent to the server. If records are sent faster than they can be delivered to the server the producer will block for max.block.ms after which it will throw an exception.
This setting should correspond roughly to the total memory the producer will use, but is not a hard bound since not all memory the producer uses is used for buffering. Some additional memory will be used for compression (if compression is enabled) as well as for maintaining in-flight requests.

Type: longDefault: 33554432Valid Values: [0,...]Importance: high
```
ref: https://kafka.apache.org/documentation/

Validated: (changed 1 config with )
`writer.kafka.producerConfig.buffer.memory=4194304`
```  
kafkaproducer.buffer-available-bytes cluster=testkafka: 4194304
kafkaproducer.buffer-total-bytes cluster=testkafka: 4194304
```


singer/src/test/java/com/pinterest/singer/writer/TestKafkaProducerManager.java
```diff
@@ -0,0 +1,67 @@
...
+    val = (long) totalMemory.get(producer);
+    assertEquals(32 * 1024 * 1024, val);
+
+    // check case with less than 1024 bytes memory
+    KafkaProducerManager.getInstance().getProducers().clear();
+    config = new KafkaProducerConfig("/var/serverset/discovery.kafka.prod",
+        Arrays.asList("localhost:9092"), "-1");
+    config.setBufferMemory(KafkaUtils.DEFAULT_PRODUCER_BUFFER_MEMORY - 1);
+    producer = KafkaProducerManager.getProducer(config);
+    assertEquals(32 * 1024 * 1024, val);
```
kabochya: val is not updated

ambud: Fixed

singer-commons/src/main/java/com/pinterest/singer/utils/KafkaUtils.java
```diff
@@ -62,6 +63,11 @@
     properties.put(ProducerConfig.BOOTSTRAP_SERVERS_CONFIG, brokerList);
     properties.put(ProducerConfig.KEY_SERIALIZER_CLASS_CONFIG, config.getKeySerializerClass());
     properties.put(ProducerConfig.VALUE_SERIALIZER_CLASS_CONFIG, config.getValueSerializerClass());
+    if (config.getBufferMemory() >= DEFAULT_PRODUCER_BUFFER_MEMORY) {
+      // make sure that there is atleast some reasonable amount of memory buffer
```
zzhhhzz: typo: at least

ambud:
> a general question: since a given KafkaProducer is created by KafkaProducerManager and shared by multiple LogStreams (SingerLogConfigs), what if their buffer.memory settings are different? in this case, do we have a deterministic buffer.memory value that is being used?

That's a great question! The short answer is the first config for a fleet will be picked and the reason is since Singer can be deployed in mix and match way on different fleets. e.g one singer config may exist on one fleet but not the other so between two fleets the only way to guarantee the memory is to have the same config today.

In a future PR, we can build a cluster config mapping config that allows us to refactor the base configurations by cluster. I have deliberately avoided that for this PR to minimize the amount of code change since we are still testing this memory reduction approach.

singer-commons/src/main/java/com/pinterest/singer/utils/KafkaUtils.java
ambud: Fixed

PR #42, opened 2020-03-19
Two changes to thrift-logger module

zzhhhzz:
1. upgrade logback to 1.1.8
2. configure thread to sleep x seconds before closing thrift loggers.

### DIFF
--- a/thrift-logger/src/main/java/com/pinterest/singer/client/BaseThriftLoggerFactory.java
+++ b/thrift-logger/src/main/java/com/pinterest/singer/client/BaseThriftLoggerFactory.java
@@ -1,12 +1,12 @@
 /**
  * Copyright 2019 Pinterest, Inc.
- * 
+ *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
  * You may obtain a copy of the License at
- * 
+ *
  *    http://www.apache.org/licenses/LICENSE-2.0
- * 
+ *
  * Unless required by applicable law or agreed to in writing, software
  * distributed under the License is distributed on an "AS IS" BASIS,
  * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
@@ -18,6 +18,7 @@ package com.pinterest.singer.client;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import java.util.Map;
 import java.util.concurrent.ConcurrentHashMap;
 
 /**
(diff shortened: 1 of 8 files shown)


thrift-logger/src/main/java/com/pinterest/singer/client/BaseThriftLoggerFactory.java
```diff
@@ -34,6 +35,16 @@
   protected ConcurrentHashMap<String, ThriftLogger> loggersByTopic
       = new ConcurrentHashMap<String, ThriftLogger>();
 
+  private int sleepInSecBeforeCloseLoggers = -1;
```
vahidhashemian: It seems this number gets converted to ms value later. Why not define this as `sleepInMsBeforeCloseLoggers`?

thrift-logger/src/main/java/com/pinterest/singer/client/BaseThriftLoggerFactory.java
```diff
@@ -93,8 +104,19 @@ private synchronized ThriftLogger getOrCreateLogger(ThriftLoggerConfig thriftLog
...
   public synchronized void shutdown() {
-    for (ThriftLogger logger : loggersByTopic.values()) {
-      logger.close();
+    try{
+      if (sleepInSecBeforeCloseLoggers > 0){
+        LOGGER.info("Before closing loggers, sleep {} seconds", sleepInSecBeforeCloseLoggers);
+        Thread.sleep(sleepInSecBeforeCloseLoggers * 1000);
+        LOGGER.info("After {} seconds, start to close loggers", sleepInSecBeforeCloseLoggers);
+      }
+    }catch(InterruptedException e){
```
HenryCaiHaiying: missing space

thrift-logger/src/main/java/com/pinterest/singer/client/BaseThriftLoggerFactory.java
```diff
@@ -93,8 +104,19 @@ private synchronized ThriftLogger getOrCreateLogger(ThriftLoggerConfig thriftLog
...
+        LOGGER.warn("Thread got interrupted", e);
+    } finally{
```
HenryCaiHaiying: missing space

thrift-logger/src/main/java/com/pinterest/singer/client/BaseThriftLoggerFactory.java
zzhhhzz: In singer-common module ( https://github.com/pinterest/singer/blob/master/singer-commons/src/main/java/com/pinterest/singer/loggingaudit/client/AuditEventKafkaSender.java#L82 ),  we have similar parameter (in seconds) to control the shutdown behavior.   I feel it is better to be consistent in naming across modules.

thrift-logger/src/main/java/com/pinterest/singer/client/BaseThriftLoggerFactory.java
zzhhhzz: fix in the update of the PR

thrift-logger/src/main/java/com/pinterest/singer/client/BaseThriftLoggerFactory.java
zzhhhzz: fix in the update of the PR

PR #43, opened 2020-03-19
Simple Mapped Thrift Reader implementation, this enables efficient batch loading

ambud:
Simple Mapped Thrift Reader implementation, this enables efficient batch loading of log data. Note that this PR only contributes the implementation without changing any critical wiring of Singer writer code making it safer to commit.

### DIFF
--- a/singer/src/main/java/com/pinterest/singer/reader/mapped/MappedFileTBinaryProtocol.java
+++ b/singer/src/main/java/com/pinterest/singer/reader/mapped/MappedFileTBinaryProtocol.java
@@ -1,0 +1,407 @@
+/**
+ * Copyright 2019 Pinterest, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.pinterest.singer.reader.mapped;
+
+import java.io.File;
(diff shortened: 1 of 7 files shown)


singer/src/main/java/com/pinterest/singer/reader/mapped/MappedThriftLogFileReader.java
```diff
@@ -0,0 +1,195 @@
...
+ * <p>
+ * This class is NOT thread-safe.
+ */
+public class MappedThriftLogFileReader implements LogFileReader {
+
+  private static final Logger LOG = LoggerFactory.getLogger(MappedThriftLogFileReader.class);
+
+  // Factory that create LogMessage thrift objects.
+  private static final class LogMessageFactory
+      implements MappedThriftReader.TBaseFactory<com.pinterest.singer.thrift.LogMessage> {
```
vahidhashemian: Does `LogMessage` need to be fully qualified here and below, since it's imported above?

singer/src/main/java/com/pinterest/singer/reader/mapped/MappedFileTBinaryProtocol.java
```diff
@@ -0,0 +1,407 @@
...
+   * @param size
+   * @throws TException
+   */
+  protected void checkAvailable(int size) throws TException {
+    if (buffer.remaining() < size) {
+      if (file.length() > buffer.capacity()) {
+        remap();
+      } else {
+        throw new TException(
+            "No more data available: " + file.length() + " vs." + buffer.capacity());
```
vahidhashemian: Might help if we identify which is which in the message (length vs. capacity).

singer/src/main/java/com/pinterest/singer/reader/mapped/MappedThriftLogFileReader.java
ambud: Fixed

singer/src/main/java/com/pinterest/singer/reader/mapped/MappedFileTBinaryProtocol.java
ambud: Fixed

PR #60, opened 2020-06-06
Add new MemoryEfficientLogStreamProcessor to reduce memory utilization

ambud:
New MemoryEfficientLogStreamProcessor to reduce memory utilization. 

**Description:**
Singer currently uses a micro-batch processing architecture which relies on in-memory buffering of data using List. This design requires substantial memory usage as all data need to be first loaded in memory instead of being directly available for streaming.

This change creates an iterator like design pattern that can be leveraged to stream data directly to the writer and let the writer create and ship micro-batches as needed. This allows only 1 LogMessage object to be buffered in memory instead of 10s or 100s (depending on configured logstream batch sizes) making Singer substantially more scalable and allowing it to ship more data without consuming more resources.

**Configurations:**
The new processor can be configured on a per LogStream basis to reduce impact and enable selective rollout.

**Test Setup:**
Using random log generation application generating thrift logs, thrift log files of 2GB were created. 
- Total logs: 2GB
- Message size: 5-8KB
- Message contents: UUID.randomUUID() + text data
- Destination: Kafka (same cluster)
- Reader Type: thrift
- Processing Batch size: 10K
- Processing Frequency: 2s

Multiple runs were performed to capture the Singer gc log data to confirm the active heap utilization. In each case Singer was started with 800MB of heap (-Xms800m -Xmx800m)

**Test Results:**
Old version of singer (regular)
Heap usage (min): 250-300MB
Heap usage (max): 600MB

New version of singer (shadow mode)
Heap usage (min): 25MB
Heap usage (max): 30MB

Utilization memory is proportional to the size of the Kafka buffer.memory which can be further optimized as needed.

### DIFF
--- a/singer/src/main/java/com/pinterest/singer/common/LogStreamWriter.java
+++ b/singer/src/main/java/com/pinterest/singer/common/LogStreamWriter.java
@@ -43,4 +43,41 @@ public interface LogStreamWriter extends Closeable {
    * @throws LogStreamWriterException when writer fails to write the LogMessages.
    */
   void writeLogMessages(List<LogMessage> messages) throws LogStreamWriterException;
+
+  /**
+   * @return if this writer implementation is committable
+   */
+  default boolean isCommittableWriter() {
+    return false;
+  }
+  
+  /**
+   * Start a new commit and run any pre-commit steps
+   * @throws LogStreamWriterException
+   */
+  default void startCommit() throws LogStreamWriterException {
+    throw new UnsupportedOperationException();
+  }
+
+  /**
+   * Send 1 LogMessage to the writer, note that writer is expected to not finalize
+   * the messages until the commit method is invoked
+   * 
+   * @param message
+   * @throws LogStreamWriterException
+   */
+  default void writeLogMessageToCommit(LogMessage message) throws LogStreamWriterException {
+    throw new UnsupportedOperationException();
+  }
+
+  /**
+   * Commit all logmessages written using {@link LogStreamWriter#writeCommittableLogMessage}
+   * 
+   * NOTE: by default this method throws UnsupportedOperationException
+   * @param numLogMessagesRead 
+   * @throws LogStreamWriterException
+   */
+  default void endCommit(int numLogMessagesRead) throws LogStreamWriterException {
+    throw new UnsupportedOperationException();
+  }
 }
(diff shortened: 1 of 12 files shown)


singer/src/main/java/com/pinterest/singer/common/LogStreamWriter.java
```diff
@@ -43,4 +43,41 @@
...
+   * @param message
+   * @throws LogStreamWriterException
+   */
+  default void writeLogMessageToCommit(LogMessage message) throws LogStreamWriterException {
+    throw new UnsupportedOperationException();
+  }
+
+  /**
+   * Commit all logmessages written using {@link LogStreamWriter#writeCommittableLogMessage}
+   *
```
zzhhhzz, line 75: what is the purpose / use case of adding these methods?

singer/src/main/java/com/pinterest/singer/common/LogStreamWriter.java
```diff
@@ -43,4 +43,41 @@
...
+  /**
+   * Commit all logmessages written using {@link LogStreamWriter#writeCommittableLogMessage}
+   * 
+   * NOTE: by default this method throws UnsupportedOperationException
+   * @param numLogMessagesRead 
+   * @throws LogStreamWriterException
+   */
+  default void endCommit(int numLogMessagesRead) throws LogStreamWriterException {
+    throw new UnsupportedOperationException();
+  }
```
zzhhhzz, line 82: what is the purpose / use case of these methods?

singer/src/main/java/com/pinterest/singer/processor/DefaultLogStreamProcessor.java
```diff
@@ -69,16 +69,16 @@
   private final String logDecider;
 
   // LogStream to be processed.
-  private final LogStream logStream;
+  protected final LogStream logStream;
```
zzhhhzz, line 72: why does the modifier changed from private to protected?

ambud:
> pls add javadoc to these new classes added. 

Added.

> overall the PR looks good. My major concern:
> (1) on a log stream using MemoryEfficientLogStreamProcessor and CommittableKafkaWriter instead of the default ones, what is the latency change? I am assuming the latency will go up.

There is no observable change to latency, shared graphs. The only key difference between the two implementations which thread runs compression code (if compression is enabled) but given the average size of batches and the cpu utilization of compression algorithm this latency isn't measurable at the singer latency level i.e. < ms 
If this is really a concern then Memory Efficiency shouldn't be enabled for such logstreams. (Note it's optional to enable this and isn't default Singer behavior)

> (2) if one log stream using the new writer class and new processor class, what is the impact in terms of latency and processing speed on other log streams using the default writer and default processor? Asking this because the logProcessorExecutor is a global resource for all the log stream.
There isn't any additional impact than the current setup, as mentioned above the new tasks perform lower amount of effort than the Default design. Duration of tasks is dependent on Kafka or any other writer that is used.

zzhhhzz:
lgtm

PR #70, opened 2020-07-29
Improve tests and refactor code for committable writers, fix env regex

ambud:
Improve tests and refactor code for committable writers, fix env regex issue and added tests.
Singer version 0.8.0.21

### DIFF
--- a/singer/src/main/java/com/pinterest/singer/writer/kafka/CommittableKafkaWriter.java
+++ b/singer/src/main/java/com/pinterest/singer/writer/kafka/CommittableKafkaWriter.java
@@ -169,12 +169,10 @@ public class CommittableKafkaWriter extends KafkaWriter {
             "host=" + HOSTNAME, "logName=" + msg.getLoggingAuditHeaders().getLogName(),
             "logStreamName=" + logName);
       }
-      // it is the index of the audited message within its bucket.
-      // note that not necessarily all messages within a bucket are being audited,
-      // thus which
-      // message within the bucket being audited should be keep track of for later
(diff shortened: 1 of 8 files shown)


singer/src/main/java/com/pinterest/singer/writer/kafka/CommittableKafkaWriter.java
```diff
@@ -290,6 +249,56 @@ public void endCommit(int numLogMessages) throws LogStreamWriterException {
...
+      OpenTsdbMetricConverter.incr(SingerMetrics.AUDIT_HEADERS_METADATA_COUNT_MATCH, 1,
+          "host=" + HOSTNAME, "logStreamName=" + logName);
+    }
+  }
+
+  private void updateWriteFailureMetrics(int numLogMessages) {
+    OpenTsdbMetricConverter.incr("singer.writer.producer_reset", 1, "topic=" + topic,
+        "host=" + HOSTNAME);
+    OpenTsdbMetricConverter.incr("singer.writer.num_kafka_messages_delivery_failure",
+        numLogMessages, "topic=" + topic, "host=" + HOSTNAME, "logname=" + logName);
```
vahidhashemian, line 284: Should these metric names also be defined in `SingerMetrics` class as constants?

ambud: Yes, but the metric name should be changed to num_messages_delivery. Once we are able to refactor the names we should be able move these to SingerMetrics

PR #71, opened 2020-07-29
Add crc features for thrift-logger and singer

elaneyc:
1. For thrift-logger, add crc feature by computing and setting the checksum field of the LogMessage
2. For singer, add crc feature in KafkaWriter.java and CommittableKafkaWriter.java, and inject the crc checksum as a header of the ProducerRecord

### DIFF
--- a/singer-commons/src/main/thrift/loggingaudit_config.thrift
+++ b/singer-commons/src/main/thrift/loggingaudit_config.thrift
@@ -56,6 +56,11 @@ struct AuditConfig{
     *  true at Merced / Merced_HR stage.
     */
     3: optional bool stopAtCurrentStage = false;
+
+    /**
+    *   flag indicates whether to skip corrupted messages at the current stage or not
+    */
+    4: optional bool skipCorruptedMessageAtCurrentStage = false;
 }
 
 
--- a/singer/src/main/java/com/pinterest/singer/writer/HeadersInjector.java
+++ b/singer/src/main/java/com/pinterest/singer/writer/HeadersInjector.java
@@ -15,7 +15,6 @@
  */
 package com.pinterest.singer.writer;
 
-import com.pinterest.singer.thrift.LogMessage;
 import org.apache.kafka.common.header.Headers;
 
 /**
@@ -26,11 +25,10 @@ import org.apache.kafka.common.header.Headers;
 public interface HeadersInjector {
 
   /**
-   * Given ProducerRecord's Headers object and LogMessage object
+   * Given ProducerRecord's Headers object and String key and byte[] value
    *
    * @return The original ProducerRecord's Headers object with some headers added.
    */
-
-   Headers addHeaders(Headers headers, LogMessage logMessage);
+   Headers addHeaders(Headers headers, String key, byte[] value);
 
 }
--- a/singer/src/main/java/com/pinterest/singer/writer/KafkaWriter.java
+++ b/singer/src/main/java/com/pinterest/singer/writer/KafkaWriter.java
@@ -15,6 +15,7 @@
  */
 package com.pinterest.singer.writer;
 
+import com.google.common.primitives.Longs;
 import com.pinterest.singer.common.LogStream;
 import com.pinterest.singer.common.LogStreamWriter;
 import com.pinterest.singer.common.errors.LogStreamWriterException;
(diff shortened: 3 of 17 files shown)


singer-commons/src/main/thrift/loggingaudit_config.thrift
```diff
@@ -56,6 +56,11 @@ struct AuditConfig{
     *  true at Merced / Merced_HR stage.
     */
     3: optional bool stopAtCurrentStage = false;
+
+    /**
+    *   flag indicates whether to skip corrupted messages at the current stage or not
+    */
+    4: optional bool skipCorruptedMessageAtCurrentStage = false;
```
zzhhhzz, line 63: since this is a new field introduced in AuditConfig,  at Singer side,  we need to parse the configuration if it is provided.  pls take a look at the method parseAuditConfig() in ConfigUtils.java

singer/src/main/java/com/pinterest/singer/writer/HeadersInjector.java
```diff
@@ -26,11 +26,10 @@
 public interface HeadersInjector {
 
   /**
-   * Given ProducerRecord's Headers object and LogMessage object
+   * Given ProducerRecord's Headers object and String key and byte[] value
```
zzhhhzz, line 29: pls remove the unused import if LogMessage is not used in this refactoring

singer/src/test/java/com/pinterest/singer/writer/kafka/TestCommittableKafkaWriter.java
```diff
@@ -292,4 +305,159 @@ public void testWriterWithHeadersInjectorEnabledWithWrongClass() throws Exceptio
...
+            new PartitionInfo("topicx", 10, new Node(1, "broker1", 9092, "us-east-1a"), null, null)));
+    when(producer.partitionsFor("topicx")).thenReturn(partitions);
+    when(producer.send(any())).thenReturn(ConcurrentUtils.constantFuture(
+            new RecordMetadata(new TopicPartition("topicx", 0), 0L, 0L, 0L, 0L, 0, 0)));
+
+    List<LogMessage> logMessages = new ArrayList<>();
+    populateLogMessagesForCorruptedMessagesTests(logMessages, partitions);
+
+    List<byte[]> corruptedMessageKeys = new ArrayList<byte[]>();
+    for (int i = NUM_EVENTS - 1; i > NUM_EVENTS - 10; i--) {
```
ambud, line 389: If we corrupting a constant number of events shouldn't that constant number be used in the assert statement?

singer/src/main/java/com/pinterest/singer/writer/KafkaWriter.java
```diff
@@ -332,13 +342,85 @@ public ExecutorService getClusterThreadPool() {
...
+    boolean shouldSkipMessage = false;
+    boolean isMessageValid = checkMessageValid(msg);
+    if (enableLoggingAudit && auditConfig.isSkipCorruptedMessageAtCurrentStage() && !isMessageValid) {
+      OpenTsdbMetricConverter.incr(SingerMetrics.NUM_CORRUPTED_MESSAGES_SKIPPED, "topic=" + topic,
+              "host=" + HOSTNAME, "logName=" + msg.getLoggingAuditHeaders().getLogName(),
+              "logStreamName=" + logName);
+      shouldSkipMessage = true;
+    }
+    if (this.headersInjector != null) {
+      injectHeadersForProducerRecord(msg, headers);
```
ambud, line 372: @zzhhhzz is there a reason for still performing header injection if the message is corrupt & will be skipped?

zzhhhzz: @ambud  yes.   When messages are decided as corrupted when the crc checksum computed by Singer is different from the value computed by thrift-logger,  we still want to injector headers as well as the messageCRC, because we can configure downstream system like Merced to detect and remove the corrupted messages.  the audit events sent out from different stages have LoggingAuditHeaders as one of its field and real-time processing will group audit events based on the LoggingAuditHeaders to decide data loss & corruption.

zzhhhzz: when shouldSkipMessage is true,  there is no need to inject headers.   should change the if conditions:    if (!shouldSkipMessage && this.headersInjector != null) ...

PR #73, opened 2020-08-06
Add a metric to measure singer startup latency

ambud:
Add a metric to measure singer startup latency.

### DIFF
--- a/singer/src/main/java/com/pinterest/singer/common/SingerSettings.java
+++ b/singer/src/main/java/com/pinterest/singer/common/SingerSettings.java
@@ -112,7 +112,7 @@ public final class SingerSettings {
              NoSuchMethodException,
       SingerLogException {
     setSingerConfig(config);
-    
+    long startTs = System.currentTimeMillis();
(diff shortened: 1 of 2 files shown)


singer/src/main/java/com/pinterest/singer/common/SingerSettings.java
```diff
@@ -216,7 +216,9 @@ public static void initialize(SingerConfig config)
       logMonitor = (LogMonitor) getInstanceMethod.invoke(null, monitorIntervalInSecs, singerConfig);
       logMonitor.start();
     }
-
+    long startupLatency = System.currentTimeMillis() - startTs;
+    Stats.setGauge(SingerMetrics.SINGER_STARTUP_LATENCY, startupLatency);
+    
     globalFsm.start();
```
zzhhhzz, line 222: why not place the latency calculation after globalFsm.start();

ambud: To prevent any future blocking calls from causing the metric to go to infinity.

PR #79, opened 2020-08-20
Add message key/value size metric in the processor

vahidhashemian:
`processor.message.key.size.bytes`
`processor.message.value.size.bytes`

Current implementation sends two metrics per message. If this could cause stress on the metrics processing side, we could easily change it to send two aggregate metrics per batch of messages.

### DIFF
--- a/singer/src/main/java/com/pinterest/singer/processor/DefaultLogStreamProcessor.java
+++ b/singer/src/main/java/com/pinterest/singer/processor/DefaultLogStreamProcessor.java
@@ -31,6 +31,7 @@ import com.pinterest.singer.thrift.LogMessage;
 import com.pinterest.singer.thrift.LogMessageAndPosition;
 import com.pinterest.singer.thrift.LogPosition;
 import com.pinterest.singer.utils.LogConfigUtils;
+import com.pinterest.singer.utils.SingerUtils;
 import com.pinterest.singer.utils.WatermarkUtils;
 
 import com.google.common.base.Preconditions;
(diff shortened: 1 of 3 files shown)


singer/src/main/java/com/pinterest/singer/processor/DefaultLogStreamProcessor.java
```diff
@@ -549,7 +550,18 @@ private void writeLogMessages(List<LogMessageAndPosition> logMessagesRead)
     }
     List<LogMessage> logMessagesToWrite = Lists.newArrayListWithExpectedSize(numMessages);
     for (LogMessageAndPosition logMessageRead : logMessagesRead) {
-      logMessagesToWrite.add(logMessageRead.getLogMessage());
+      LogMessage logMessage = logMessageRead.getLogMessage();
+      logMessagesToWrite.add(logMessage);
+      OpenTsdbMetricConverter.gauge(
+              SingerMetrics.PROCESSOR_MESSAGE_KEY_SIZE_BYTES,
+              logMessage.getKey().length,
```
ambud, line 557: This will throw an NPE when key is null

singer/src/main/java/com/pinterest/singer/processor/MemoryEfficientLogStreamProcessor.java
```diff
@@ -96,7 +99,18 @@ protected int processLogMessageBatch() throws IOException, LogStreamWriterExcept
         // because there is some data to read we need to prepare the commit
         writer.startCommit();
       }
-      writer.writeLogMessageToCommit(message.getLogMessage());
+      LogMessage logMessage = message.getLogMessage();
+      OpenTsdbMetricConverter.gauge(
+              SingerMetrics.PROCESSOR_MESSAGE_KEY_SIZE_BYTES,
+              logMessage.getKey().length,
```
ambud, line 105: Potential NPE when key is null

ambud: We should also add a test to processor to add null keys to capture and validate the null tests.

singer/src/main/java/com/pinterest/singer/processor/DefaultLogStreamProcessor.java
vahidhashemian: Good point. I think we should still send a metric if key is null so we can correlated values for key/value pairs. Would it make sense to send a fixed number like `-1` when key is null?

singer/src/main/java/com/pinterest/singer/processor/MemoryEfficientLogStreamProcessor.java
vahidhashemian: Do we want unit test for the metrics piece? Or general unit tests to cover null keys?

ambud: General processor unit test to cover null keys

singer/src/main/java/com/pinterest/singer/processor/DefaultLogStreamProcessor.java
ambud: If the key is null the bytes should be 0. -1 will provide an incorrect signal.

PR #80, opened 2020-08-20
use default partition instead of random partition in Audit client library

zzhhhzz:
Instead of choose random partition, AuditEventKafkaSender chooses default partition of the audit_event topic to send audit events. This reduces the number of TCP connections from audit client to Kafka cluster hosting the auditi_event topic.

Default partition is updated if it is in badPartitions set.


singer-commons/src/main/java/com/pinterest/singer/loggingaudit/client/AuditEventKafkaSender.java
```diff
@@ -216,6 +212,33 @@ private void refreshPartitionIfNeeded() {
     }
   }
 
+  private void resetDefaultPartitionIfNeeded() {
+    if (partitionInfoList.size() == 0) {
+      defaultPartition = 0;
```
ambud, line 217: If partition count is 0 then defaultPartition shouldn't be set to a valid Kafka partition id.

singer-commons/src/main/java/com/pinterest/singer/loggingaudit/client/AuditEventKafkaSender.java
```diff
@@ -165,6 +165,13 @@
 
   private Map<LoggingAuditHeaders, Integer> eventTriedCount = new ConcurrentHashMap<>();
 
+  /**
+   *  Default partition of the audit_event topic used to receive audit events. Default partition
+   *  will be reset in resetDefaultPartitionIfNeeded() method. This reduces the number of TCP
+   *  connections from audit client to the Kafka Cluster hosting the audit_event topic.
+   */
+  private int defaultPartition = -1;
```
ambud, line 173: Should this be called `currentPartitionId` instead of defaultPartition?

singer-commons/src/main/java/com/pinterest/singer/loggingaudit/client/AuditEventKafkaSender.java
```diff
@@ -232,13 +255,13 @@ public void run() {
     while (!cancelled.get()) {
       try {
         refreshPartitionIfNeeded();
+        resetDefaultPartitionIfNeeded();
```
ambud, line 258: Shouldn't resetDefaultPartition be called within refreshPartition? Partition Id setting can be done in one method instead of two.
Also setting valid partitionId should only be done during refresh interval checks for efficiency reasons else `resetDefaultPartitionIfNeeded` method will unnecessarily run the same logic again.

singer-commons/src/main/java/com/pinterest/singer/loggingaudit/client/AuditEventKafkaSender.java
zzhhhzz: sure. currentPartitionId sounds better.  will update this

singer-commons/src/main/java/com/pinterest/singer/loggingaudit/client/AuditEventKafkaSender.java
zzhhhzz: the logic behind this is when partitionInfoList.size() is zero,  this means the audit client couldn't get the topic metadata from the Kafka cluster at this moment. metadata being not available could be due to temporal network glitch or Kafka cluster outage.   if we set defaultPartition to be invalid id such as -1,  then the following send() method will ALWAYS fail.  Setting to 0 will increase the chance of successfully sending audit events to Kafka.

ambud: defaultPartition cannot be 0 if partition size is 0 as that logic is invalid.

zzhhhzz:
updated the PR

zzhhhzz:
resetCurrentPartitionIdIfNeeded() is called within refreshPartitionIfNeeded().  it will update currentPartitionId as long as it is in badPartitions set.  If there is no available partitions, i.e. partitionInfoList is empty, then currentPartitionId will be set as -1.

PR #86, opened 2020-08-25
Cleanup unnecessary Singer dependencies to make binary more efficient

ambud:
Cleanup unnecessary Singer dependencies to make binary more efficient. Singer 0.8.0.30

- Marking Pulsar dependencies as provided so anyone deploying Singer with Pulsar can optionally add those; this saves over 15MB deployment size
- Removing unnecessary dependency on Kafka server


ambud:
Will re-initiate a PR

PR #91, opened 2020-09-10
Fix bug for missing cluster name config. Singer 0.8.0.34

ambud:
Fix bug for missing cluster name config. Singer 0.8.0.34

### DIFF
--- a/singer/src/main/java/com/pinterest/singer/utils/LogConfigUtils.java
+++ b/singer/src/main/java/com/pinterest/singer/utils/LogConfigUtils.java
@@ -272,6 +272,7 @@ public class LogConfigUtils {
     }
 
     String clustername = configuration.getString("cluster");
+    config.setCluster(clustername);
     String environment = configuration.getString("environment");
     if (!Arrays.asList("dev", "test", "prod").contains(environment)) {
       throw new ConfigurationException("Invalid environment name");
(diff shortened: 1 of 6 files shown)


singer/src/main/java/com/pinterest/singer/utils/LogConfigUtils.java
```diff
@@ -272,16 +272,14 @@ public static MemqWriterConfig parseMemqWriterConfig(SubsetConfiguration configu
...
     String environment = configuration.getString("environment");
     if (!Arrays.asList("dev", "test", "prod").contains(environment)) {
       throw new ConfigurationException("Invalid environment name");
     }
     // e.g. discovery.memq.dev.prototype.prod_rich_data
     String serverSetFilePath = "/var/serverset/discovery.memq." + environment + "." + clustername
         + ".prod_rich_data";
-    if (!new File(serverSetFilePath).exists()) {
-      throw new ConfigurationException("Serverset doesn't exist:" + serverSetFilePath);
-    }
```
vahidhashemian, line 284: Is this no longer required?

vahidhashemian (approved):

singer/src/main/java/com/pinterest/singer/utils/LogConfigUtils.java
ambud: The validation is moved into the writer

PR #100, opened 2020-10-13
Multiple improvements to thrift-logger module and singer-commons module

zzhhhzz:
1. separate injecting LoggingAuditHeaders / checksum from sending out audit events:
(1) inject LoggingAuditHeaders / checksum for every message
(2) if tracked field of LoggingAuditHeaders is set true, audit event will be sent out

2. add tracked field and skipped field to loggingaudit.thrift and support passing skipped field
when calling audit method of LoggingAuditClient

3. add enableDeserialization field and messageClass field to AuditConfig struct. it will be used later
to perform thrift deserialization if configured in order to check if a message is valid or not.

### DIFF
--- a/singer-commons/src/main/java/com/pinterest/singer/loggingaudit/client/LoggingAuditClient.java
+++ b/singer-commons/src/main/java/com/pinterest/singer/loggingaudit/client/LoggingAuditClient.java
@@ -153,6 +153,20 @@ public class LoggingAuditClient {
   /**
    *  create and enqueue a LoggingAuditEvent if TopicAuditConfig exists for this topic/logName,
    *  enqueueEnabled is true and there is capacity available in the queue.
+   * @param loggingAuditHeaders
+   * @param messageValid
+   * @param messageAcknowledgedTimestamp
+   * @param messageSkipped
+   */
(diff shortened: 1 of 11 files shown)


singer-commons/src/main/java/com/pinterest/singer/loggingaudit/client/LoggingAuditClient.java
```diff
@@ -150,6 +150,20 @@ public void audit(String loggingAuditName, LoggingAuditHeaders loggingAuditHeade
         "");
   }
 
+  /**
+   *  create and enqueue a LoggingAuditEvent if TopicAuditConfig exists for this topic/logName,
+   *  enqueueEnabled is true and there is capacity available in the queue.
+   * @param loggingAuditHeaders
+   * @param messageValid
+   * @param messageAcknowledgedTimestamp
+   * @param skipped
```
ambud, line 159: Could you please rename this variable to make it more descriptive?
`messageSkipped`

It's not clear whether it's referring to a configuration or a state of the message.

ambud (changes requested):

ambud (approved):

PR #112, opened 2021-02-18
Add defaults for buffer size and max message size

ambud:
Add defaults for buffer size and max message size to avoid requiring these values to be specified for every config

### DIFF
--- a/singer/src/main/java/com/pinterest/singer/utils/LogConfigUtils.java
+++ b/singer/src/main/java/com/pinterest/singer/utils/LogConfigUtils.java
@@ -950,17 +950,16 @@ public class LogConfigUtils {
   }
 
   private static ThriftReaderConfig parseThriftReaderConfig(AbstractConfiguration thriftReaderConfiguration) {
-    thriftReaderConfiguration.setThrowExceptionOnMissing(true);
-    int readerBufferSize = thriftReaderConfiguration.getInt("readerBufferSize");
-    int maxMessageSize = thriftReaderConfiguration.getInt("maxMessageSize");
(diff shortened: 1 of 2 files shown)


singer/src/main/java/com/pinterest/singer/utils/LogConfigUtils.java
```diff
@@ -950,9 +950,8 @@ private static LogStreamReaderConfig parseLogStreamReaderConfig(AbstractConfigur
   }
 
   private static ThriftReaderConfig parseThriftReaderConfig(AbstractConfiguration thriftReaderConfiguration) {
-    thriftReaderConfiguration.setThrowExceptionOnMissing(true);
-    int readerBufferSize = thriftReaderConfiguration.getInt("readerBufferSize");
-    int maxMessageSize = thriftReaderConfiguration.getInt("maxMessageSize");
+    int readerBufferSize = thriftReaderConfiguration.getInt("readerBufferSize", 10240);
+    int maxMessageSize = thriftReaderConfiguration.getInt("maxMessageSize", 100000);
```
vahidhashemian, line 954: Nit: Defining defaults like these in one place (e.g. using constants) and referencing them here would make finding those defaults easier in the future.

vahidhashemian (approved):

singer/src/main/java/com/pinterest/singer/utils/LogConfigUtils.java
ambud: Good catch, was too focused on the bug fix so missed this. Fixed now.

vahidhashemian (approved):

PR #116, opened 2021-03-04
Add FirstByteIntPartitioner

samhsiung:
Add a partitioner that allows for manually specifying a partition to write to. This is done by composing the first four bytes of the object as a byte array into an integer value and then applying the module of the number of partitions.

Screen Shot 2021-03-04 at 9 20 39 AM


samhsiung:
Abandoning

PR #132, opened 2021-06-30
Rule-based overrides for log configs

kabochya:
Add capability of doing rule-based (currently only exact match) override of log configs
Can be done by setting `configOverrideDir` in `singer.properties` and have property files in that directory with `match.config.name` set to the config name to be matched, and `match.config.value` to the config value; add `override.` prefix to the configs that need to be overriden. Files will be sorted in alphabetical order and later overrides will have higher precedence over earlier ones.

Example:
In the `singer.properties` file:
```
...
singer.configOverrideDir=/path/to/override/dir
...
```

In `path/to/override/dir/100-override-kafka-buffer-memory.properties`
```
match.config.name=writer.type
match.config.value=kafka

override.writer.kafka.producerConfig.buffer.memory=100000
```

in `path/to/override/dir/101-override-kafka08-buffer-memory.properties`
```
match.config.name=writer.type
match.config.value=kafka08

override.writer.kafka08.producerConfig.buffer.memory=100000
```

This setup will match all the log configs with `writer.type=kafka` or `writer.type=kafka08` and override the `writer.kafka.producerConfig.buffer.memory` config to 100000 for all of the matching property files.

### DIFF
--- a/singer-commons/src/main/thrift/config.thrift
+++ b/singer-commons/src/main/thrift/config.thrift
@@ -359,4 +359,10 @@ struct SingerConfig {
    */
   23: optional i32 statsPusherFrequencyInSeconds = 180;
 
+  /**
+   * Config override directory
+   */
+
+  24: optional string configOverrideDir;
+
 }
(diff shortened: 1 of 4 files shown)


vahidhashemian:
Can you add an example in the description to show the impact of this change?

ambud (approved):
lgtm

PR #134, opened 2021-07-21
Refactor CommittableKafkaWriter to use non-blocking request handling

kabochya:
tl; dr: Replaced `clusterThreadPool` usage by using CompletableFutures to implement non-blocking kafka request handling

In this diff, we remove the usage of clusterThreadPool in `CommitableKafkaWriter`. Previously, the `clusterThreadPool` was allocated per cluster and used for tracking metrics of the `Future<RecordMetadata>`. This caused a lot of threads to be used if Singer is publishing to a lot of clusters. Since these threads are used only for tracking, we found out that they could be eliminated and instead refactored into `CompletableFuture`s that can be executed asynchronously with the shared common `ForkJoinPool` as long as they remain non-blocking. 
We also leverage the callback in `send` to achieve this:
```
    CompletableFuture<RecordMetadata> future = new CompletableFuture<>();
    committableProducer.send(keyedMessage, (recordMetadata, exception) -> {
      if (exception != null) {
        future.completeExceptionally(exception);
      } else {
        future.complete(recordMetadata);
      }
    });
```
 After this diff, it is expected to drop the number of threads created by Singer and significantly reduce the native memory allocated for multi-cluster-publishing Singer instances.

### DIFF
--- a/singer/src/main/java/com/pinterest/singer/writer/kafka/CommittableKafkaWriter.java
+++ b/singer/src/main/java/com/pinterest/singer/writer/kafka/CommittableKafkaWriter.java
@@ -17,15 +17,21 @@ package com.pinterest.singer.writer.kafka;
 
 import java.nio.ByteBuffer;
 import java.util.ArrayList;
-import java.util.Collections;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Map.Entry;
-import java.util.concurrent.Callable;
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.CompletionException;
+import java.util.concurrent.ExecutionException;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Future;
+import java.util.concurrent.ScheduledExecutorService;
+import java.util.concurrent.ScheduledThreadPoolExecutor;
 import java.util.concurrent.TimeUnit;
+import java.util.concurrent.TimeoutException;
+import java.util.function.Function;
+import java.util.stream.Collectors;
 
 import org.apache.kafka.clients.producer.KafkaProducer;
 import org.apache.kafka.clients.producer.ProducerRecord;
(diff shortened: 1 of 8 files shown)


singer/src/main/java/com/pinterest/singer/writer/kafka/CommittableKafkaWriter.java
```diff
@@ -138,20 +148,20 @@ public void startCommit() throws LogStreamWriterException {
       }
     }
 
-    commitableBuckets = new HashMap<>();
-    committableMapOfTrackedMessageMaps = new HashMap<>();
-    committableMapOfInvalidMessageMaps = new HashMap<>();
-    committableMapOfOriginalIndexWithinBucket = new HashMap<>();
-
+    committableBuckets.clear();
```
ambud, line 151: In my past experience it's better to create a new data structure from GC cleanliness stand point.

singer/src/main/java/com/pinterest/singer/writer/kafka/CommittableKafkaWriter.java
```diff
@@ -138,20 +148,20 @@ public void startCommit() throws LogStreamWriterException {
...
-    for (int i = 0; i < committableValidPartitions.size(); i++) {
+    for (PartitionInfo partitionInfo : committableValidPartitions) {
       // for each partitionId, there is a corresponding bucket in buckets and a
       // corresponding headersMap in mapOfHeadersMaps.
-      PartitionInfo partitionInfo = committableValidPartitions.get(i);
       int partitionId = partitionInfo.partition();
-      commitableBuckets.put(partitionId, new KafkaWritingTaskFuture(partitionInfo));
-      committableMapOfTrackedMessageMaps.put(partitionId, new HashMap<Integer, LoggingAuditHeaders>());
-      committableMapOfInvalidMessageMaps.put(partitionId, new HashMap<Integer, LoggingAuditHeaders>());
+      committableBuckets.put(partitionId, new KafkaWritingTaskFuture(partitionInfo));
```
ambud, line 160: Why was there an indentation update?

ambud (commented):
Added initial comments

ambud:
Could you please add a detail description of what this PR is about and how it improves singer?

singer/src/main/java/com/pinterest/singer/writer/kafka/CommittableKafkaWriter.java
kabochya: It was actually just fixing the `commitableBuckets` typo to `committableBuckets`

ambud:
Could you also include a version update?

ambud (approved):

PR #137, opened 2021-09-07
Add instance level sampling

kabochya:
This is to add another type of sampling besides the existing message-level sampling. Instance level sampling allows a logstream to be enabled/disabled by a percentage of the fleet, rather than a percentage of the messages per host. This allows canary-style rollouts of logs. Another minor improvement is logstreams that have a zero-value decider will be removed from the actively monitored logstreams, thus stop getting tracked in the stuck logstream monitor.

### DIFF
--- a/singer/src/main/java/com/pinterest/singer/monitor/DefaultLogMonitor.java
+++ b/singer/src/main/java/com/pinterest/singer/monitor/DefaultLogMonitor.java
@@ -24,6 +24,7 @@ import com.pinterest.singer.common.errors.LogStreamReaderException;
 import com.pinterest.singer.common.LogStreamWriter;
 import com.pinterest.singer.common.errors.LogStreamWriterException;
 import com.pinterest.singer.common.SingerMetrics;
+import com.pinterest.singer.config.Decider;
 import com.pinterest.singer.metrics.OpenTsdbMetricConverter;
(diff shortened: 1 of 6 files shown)


singer/src/main/java/com/pinterest/singer/monitor/DefaultLogMonitor.java
```diff
@@ -184,6 +195,28 @@ public static LogMonitor getInstance(int monitorIntervalInSecs,
...
+    SingerLogConfig logConfig = logStream.getSingerLog().getSingerLogConfig();
+    String decider = logConfig.getLogDecider();
+    int deciderValue = Decider.getInstance().getDeciderValue(decider);
+    if (deciderValue == -1) {
+      // logstream is considered active if the decider doesn't exist
+      return false;
+    } else if (deciderValue == 0) {
+      // decider is 0
+      return true;
+    } else if (logConfig.getLogStreamProcessorConfig().getDeciderBasedSampling().equals(SamplingType.INSTANCE)) {
```
ambud, line 211: Line 205 and 211 should be combined since they are dependent on each other from a logical standpoint. This is help make the code more readable.

ambud (changes requested):

ambud (approved):

PR #161, opened 2022-02-02
Initial implementation of AdminServer with draining command

kabochya:
This commit adds an AdminServer module that will allow admin operations to be performed via a Unix domain socket (UDS). UDS server implementation is backed by the junixsocket dependency. Besides file permissions, there is also a simple pid based allowlist for commands. 
The only command provided now is the `stop` command, which is will allow Singer to drain logs before getting ready to be terminated.
A simple way to perform this operation from shell would be to use `nc` (nc-freebsd) with the following command:
`echo "stop" | nc -U /tmp/singer/admin.sock`
The command will block until all logstreams are drained, or Singer is terminated. The command will send back the message `done` to the connection when the draining is complete, signaling the script that Singer is ready to be terminated. If the socket is lost during the draining, Singer will terminate itself.

### DIFF
--- a/singer/pom.xml
+++ b/singer/pom.xml
@@ -7,7 +7,7 @@
     <parent>
         <groupId>com.pinterest.singer</groupId>
         <artifactId>singer-package</artifactId>
-        <version>0.8.0.63</version>
+        <version>0.8.0.64</version>
         <relativePath>../pom.xml</relativePath>
     </parent>
     <licenses>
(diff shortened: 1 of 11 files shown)


singer/pom.xml
```diff
@@ -200,6 +200,17 @@
 			<artifactId>metrics-core</artifactId>
 			<version>4.0.2</version>
 		</dependency>
+        <dependency>
+            <groupId>com.kohlschutter.junixsocket</groupId>
+            <artifactId>junixsocket-core</artifactId>
+            <version>2.4.0</version>
+            <type>pom</type>
```
ambud, line 207: Why are we using POM dependency?

kabochya: That was what was provided in https://kohlschutter.github.io/junixsocket/dependency.html
It seems that this dependency is a parent pom.

ambud (approved):

PR #163, opened 2022-02-10
Make partitioner configurable and add FixedPartitionsPartitioner

kabochya:
Allow configs to be passed to the partitioner via `writer.kafka.producerConfig.partitioner.<config>` keys, and add a partitioner that allows fixed number of partitions to be written to (extension of `SinglePartitionPartitioner`)

### DIFF
--- a/singer/src/main/java/com/pinterest/singer/writer/partitioners/FixedPartitionsPartitioner.java
+++ b/singer/src/main/java/com/pinterest/singer/writer/partitioners/FixedPartitionsPartitioner.java
@@ -1,0 +1,77 @@
+/**
+ * Copyright 2019 Pinterest, Inc.
+ * 
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ * 
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ * 
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.pinterest.singer.writer.partitioners;
+
+import com.pinterest.singer.writer.KafkaMessagePartitioner;
+
+import org.apache.kafka.common.PartitionInfo;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.Random;
+import java.util.concurrent.ThreadLocalRandom;
+import java.util.stream.Collectors;
+
+/**
+ * This partition is used to group a batch of messages into one of fixed number of partitions
+ * This will potentially help out with both the compression ratio and connection count.
+ * The number of partitions should be chosen as a multiplier for a reasonable producer-to-partition ratio.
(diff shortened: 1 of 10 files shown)


singer/src/main/java/com/pinterest/singer/writer/partitioners/FixedPartitionsPartitioner.java
```diff
@@ -0,0 +1,67 @@
...
+
+/**
+ * This partition is used to group a batch of messages into one of at most three random partitions
+ */
+public class FixedPartitionsPartitioner implements KafkaMessagePartitioner {
+  public static final String PARTITIONER_PARTITION_COUNT_KEY = "partitioner.partition_count";
+  private static final Logger logger = LoggerFactory.getLogger(FixedPartitionsPartitioner.class);
+
+  private final Map<Integer, List<Integer>> partitionMap = new HashMap<>();
+  private final Random random = new Random(System.currentTimeMillis());
```
ambud, line 39: Can we change to ThreadLocalRandom?

singer/src/main/java/com/pinterest/singer/writer/partitioners/FixedPartitionsPartitioner.java
```diff
@@ -0,0 +1,67 @@
...
+  @Override
+  public void configure(Map<String, String> partitionerConfig) {
+    KafkaMessagePartitioner.super.configure(partitionerConfig);
+    partitionCount = Integer.parseInt(partitionerConfig.getOrDefault(PARTITIONER_PARTITION_COUNT_KEY, "3"));
+    logger.info("Partitioner configured to partition to at most " + partitionCount + " partitions");
+  }
+
+  public int partition(Object object, List<PartitionInfo> partitions) {
+    int numOfPartitions = partitions.size();
+    if (!partitionMap.containsKey(numOfPartitions)){
```
ambud, line 52: Probably worth adding a comment as to why initializations are done?

ambud (changes requested):

singer/src/main/java/com/pinterest/singer/writer/partitioners/FixedPartitionsPartitioner.java
```diff
@@ -0,0 +1,67 @@
...
+
+  private final Map<Integer, List<Integer>> partitionMap = new HashMap<>();
+  private final Random random = new Random(System.currentTimeMillis());
+  private int partitionCount = 3;
+
+
+  @Override
+  public void configure(Map<String, String> partitionerConfig) {
+    KafkaMessagePartitioner.super.configure(partitionerConfig);
+    partitionCount = Integer.parseInt(partitionerConfig.getOrDefault(PARTITIONER_PARTITION_COUNT_KEY, "3"));
```
vahidhashemian, line 46: What's the reasoning for picking `3` as the default? I was thinking maybe we can use `1` as default so it matches the current single partition partitioner. This way we can use this new partitioner as the default for fixed partitioning and deprecate single partition partitioner going forward.

singer/src/test/java/com/pinterest/singer/writer/partitioners/TestFixedNumberPartitionPartitioner.java
```diff
@@ -0,0 +1,74 @@
...
+import java.util.Arrays;
+import java.util.HashMap;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+
+public class TestFixedNumberPartitionPartitioner {
+  @Test
+  public void TestFixedPartitionsPartitioner() {
```
ambud, line 20: Could you please change it to camel case?

ambud (changes requested):

ambud (approved):

PR #238, opened 2022-09-08
Remove redundant flush() in endCommit() to fix latency issue

jeffxiang:
This is a fix for a suspicion on why the timer isn’t kicking in in Singer endCommit. KafkaProducer’s flush() call is blocking, and when there’s an issue with flushing (e.g. broker is dead or SSL issues, etc.) this call stalls, resulting in simply the producer expiring the records in the batch but the call never actually unblocks. this can result in the latency going up without the rest of the endCommit() method being invoked, which means 1) the futures don’t complete, either normally or exceptionally, and 2) the timer isn’t ever being instantiated.
the reason that the batch records can be expired without the flush returning is because the expiration logic happens in a separate Sender thread in the Producer internals.


singer/pom.xml
```diff
@@ -190,11 +190,17 @@
             <version>1.12.248</version>
         </dependency>
         <dependency>
-		    <groupId>com.salesforce.kafka.test</groupId>
-		    <artifactId>kafka-junit4</artifactId>
-		    <version>3.2.0</version>
-		    <scope>test</scope>
-		</dependency>
+            <groupId>com.salesforce.kafka.test</groupId>
```
ambud, line 193: This dependency is under BSD license whereas Singer is Apache license

jeffxiang: is there a workaround for this? we use this for the unit test simulating broker/broker down.

vahidhashemian: How come the dependency already exists in the repo? Did we miss this earlier?

vahidhashemian (approved):

PR #314, opened 2023-04-25
Add K8s uid support for pod log directories

jfzunigac:


singer/src/main/java/com/pinterest/singer/kubernetes/KubeService.java
```diff
@@ -267,11 +267,22 @@ public Set<String> fetchPodNamesFromMetadata() throws IOException {
...
+                // pod namespace
                 String namespace = metadata.get("namespace").getAsString();
-                podNames.add(namespace + "_" + name);
+                // pod uid
+                String podUid = metadata.get("uid").getAsString();
+
+                // coexist of 2 format: namespace_podname or namespace_podname_uid
+                String formatOne = namespace + "_" + name;
+                String formatTwo = namespace + "_" + name + "_" + podUid;
+                if (new File(podLogDirectory + '/' + formatOne).exists()) {
```
vahidhashemian, line 281: Will there be a scenario where a particular log stream is moving from first format to the second on the same host?

vahidhashemian (changes requested):
Please add a summary for this change.

singer/src/main/java/com/pinterest/singer/kubernetes/KubeService.java
jfzunigac: Not sure if that could be the case, but K8s team assured that two formats with the same namespace & name will not coexist in the same host. I think even if one format exists and then the other one surges, Singer will process the logs in the first format's directory and delete it before moving on since pod directory discovery is done by watching directories

jfzunigac: We could probably use a config based implementation for open source

PR #343, opened 2023-09-07
Add AZ Header to Kafka Logs

e-v-nguyen:
- Tested with unit tests to ensure "availabilityZone: "us-east-1a" shows up in log msgs to kafka.
mvn test -pl singer -am passes
- Tested in devapp w/ thrift logs and text logs, and each of those with and without EC2 env set. 
If EC2 env is set, "availabilityZone: "us-east-1a". 
Else env is not set, "availabilityZone: "n/a".


jfzunigac:
LGTM, can you also bump the version to `0.8.0.77`?

jfzunigac (approved):

PR #377, opened 2024-01-03
Add fleet level disable decider logic

jfzunigac:
Disable a logstream if fleet level disable decider is set to 100, otherwise continue with regular execution. This decider needs to be in the format: `singer_disable_<logstream_name>___<hostname_prefix>___decider` where all non non-alphanumeric charcters in `logstream_name` are replaced by underscores. Also introduce a utility function to retrieve hostname prefix list.

### DIFF
--- a/singer/src/main/java/com/pinterest/singer/processor/DefaultLogStreamProcessor.java
+++ b/singer/src/main/java/com/pinterest/singer/processor/DefaultLogStreamProcessor.java
@@ -69,7 +69,7 @@ public class DefaultLogStreamProcessor implements LogStreamProcessor, Runnable {
   private final String logDecider;
 
   // Decider used in conjunction with logDecider to disable the logstream at a fleet level
-  private final String logstreamNameForDecider;
+  private final List<String> disableDeciders;
 
   // LogStream to be processed.
   protected final LogStream logStream;
@@ -168,8 +168,9 @@ public class DefaultLogStreamProcessor implements LogStreamProcessor, Runnable {
     this.exceedTimeSliceLimit = false;
     this.lastModificationTimeProcessed = new AtomicLong(-1);
     this.lastCompletedCycleTime = new AtomicLong(-1);
-    this.logstreamNameForDecider =
-        logStream.getSingerLog().getSingerLogConfig().getName().replaceAll("[^a-zA-Z0-9]", "_");
+    this.disableDeciders =
+        Decider.getInstance().generateDisableDeciders(
+            this.logStream.getSingerLog().getSingerLogConfig().getName());
   }
 
   @Override
@@ -258,10 +259,15 @@ public class DefaultLogStreamProcessor implements LogStreamProcessor, Runnable {
       if (map.containsKey(logDecider)) {
         result = map.get(logDecider) != 0;
       }
-      String disableDecider = Decider.getInstance().getDisableDecider(logstreamNameForDecider);
-      if (result && disableDecider != null && map.get(disableDecider) == 100) {
-        LOG.info("Disabling log stream {} because fleet disable decider is set to 100", logStream.getLogStreamName());
-        result = false;
+      if (result && disableDeciders != null) {
+        for (String disableDecider : disableDeciders) {
+          if (map.containsKey(disableDecider) && map.get(disableDecider) == 100) {
+            LOG.info("Disabling log stream {} because fleet disable decider {} is set to 100",
+                logStream.getLogStreamName(), disableDecider);
+            result = false;
+            break;
+          }
+        }
       }
     }
     return result;
--- a/singer/src/main/java/com/pinterest/singer/processor/DefaultLogStreamProcessor.java
+++ b/singer/src/main/java/com/pinterest/singer/processor/DefaultLogStreamProcessor.java
@@ -248,7 +248,7 @@ public class DefaultLogStreamProcessor implements LogStreamProcessor, Runnable {
   /**
    * If the decider is not set, this method will return true.
    * If a decider is set, return false when the decider's value is 0
-   * and disable decider's (if exists) value is 100.
+   * or disable decider's (if exists) value is 100.
    *
    * @return true or false.
    */
--- a/singer/src/main/java/com/pinterest/singer/processor/DefaultLogStreamProcessor.java
+++ b/singer/src/main/java/com/pinterest/singer/processor/DefaultLogStreamProcessor.java
@@ -68,7 +68,7 @@ public class DefaultLogStreamProcessor implements LogStreamProcessor, Runnable {
   // Decider for the log stream.
   private final String logDecider;
 
-  // Valid deciders tha can be used in conjunction with logDecider to disable the logstream at a fleet level
+  // Valid deciders that can be used in conjunction with logDecider to disable the logstream at a fleet level
   private final List<String> disableDeciders;
 
   // LogStream to be processed.
@@ -247,8 +247,8 @@ public class DefaultLogStreamProcessor implements LogStreamProcessor, Runnable {
 
   /**
    * If the decider is not set, this method will return true.
-   * If a decider is set, return false when the decider's value is 0 or when the
-   * decider's value is != 0 and disable decider's (if exists) value is 100.
+   * If a decider is set, return false when the decider's value is 0
+   * and disable decider's (if exists) value is 100.
    *
    * @return true or false.
    */
--- a/singer/src/main/java/com/pinterest/singer/processor/DefaultLogStreamProcessor.java
+++ b/singer/src/main/java/com/pinterest/singer/processor/DefaultLogStreamProcessor.java
@@ -265,7 +265,7 @@ public class DefaultLogStreamProcessor implements LogStreamProcessor, Runnable {
             LOG.info("Disabling log stream {} because fleet disable decider {} is set to 100",
                 logStream.getLogStreamName(), disableDecider);
             OpenTsdbMetricConverter.gauge(
-                "singer.processor.disable_decider_active", 1, "log=" + logStream.getSingerLog().getLogName());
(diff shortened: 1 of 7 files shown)


vahidhashemian (approved):

singer/src/main/java/com/pinterest/singer/processor/DefaultLogStreamProcessor.java
```diff
@@ -68,6 +68,9 @@ public class DefaultLogStreamProcessor implements LogStreamProcessor, Runnable {
   // Decider for the log stream.
   private final String logDecider;
 
+  // Decider used in conjunction with logDecider to disable the logstream at a fleet level
```
vahidhashemian, line 71: nit: "Decider used" -> "Valid deciders that can be used"
?

singer/src/main/java/com/pinterest/singer/processor/DefaultLogStreamProcessor.java
```diff
@@ -241,7 +247,8 @@ public long processLogStream() throws LogStreamProcessorException {
 
   /**
    * If the decider is not set, this method will return true.
-   * If a decider is set, only return false when the decider's value is 0.
+   * If a decider is set, only return false when the decider's value is 0 and disable decider's
+   * (if exists) value is 100.
```
vahidhashemian, line 251: Is it "and" or "or"? I.e. do both conditions need to hold? Or either?

singer/src/main/java/com/pinterest/singer/processor/DefaultLogStreamProcessor.java
```diff
@@ -252,6 +259,16 @@ boolean isLoggingAllowedByDecider() {
       if (map.containsKey(logDecider)) {
         result = map.get(logDecider) != 0;
       }
+      if (result && disableDeciders != null) {
+        for (String disableDecider : disableDeciders) {
+          if (map.containsKey(disableDecider) && map.get(disableDecider) == 100) {
+            LOG.info("Disabling log stream {} because fleet disable decider {} is set to 100",
```
vahidhashemian, line 265: For tracking purposes would it help if we emit a metric here (maybe only the first time it is detected or when it changes so we don't overwhelm metric agent)?

singer/src/main/java/com/pinterest/singer/utils/SingerUtils.java
```diff
@@ -91,6 +95,25 @@ public static String getHostname() {
     return hostName;
   }
 
+  /***
+   * Gradually builds substrings from hostname separated by dashes
+   * will return hostname if hostname does not contain dashes
+   *
```
vahidhashemian, line 101: nit: An example would help to show what are considered hostname prefixes.

Also, considering the generic use case, should we somehow allow custom (configurable) separators (other than just `-`)?

singer/src/main/java/com/pinterest/singer/processor/DefaultLogStreamProcessor.java
jfzunigac: Yes both conditions need to hold if an applicable disable decider exists for the host

singer/src/main/java/com/pinterest/singer/processor/DefaultLogStreamProcessor.java
jfzunigac: We can, I can probably add it right before we return here so that its emitting the metric for as long as the disable decider is turned on (which ideally shouldn't be for long)

singer/src/main/java/com/pinterest/singer/utils/SingerUtils.java
jfzunigac: Yes, what do you think about setting a configurable regex pattern instead?

singer/src/main/java/com/pinterest/singer/processor/DefaultLogStreamProcessor.java
vahidhashemian: Am I missing something here?

Logging is allowed (return true) if
-  logstream decider value is non-zero, and
- disable deccider value is not 100.

Then, logging is disallowed (return false) if (logical negation of "a and b" is "~a or ~b")
- logstream decider value is 0, or
- disable decider value is 100

singer/src/main/java/com/pinterest/singer/processor/DefaultLogStreamProcessor.java
vahidhashemian: Did you see this one? ^

singer/src/main/java/com/pinterest/singer/utils/SingerUtils.java
vahidhashemian: That would be most flexible way of supporting this. Do you have an example?

singer/src/main/java/com/pinterest/singer/processor/DefaultLogStreamProcessor.java
jfzunigac: Oh sorry yes I misread your comment, I mean to convey that disable decider will only be considered if logsteam decider is non-zero. Let me change this comment

singer/src/main/java/com/pinterest/singer/utils/SingerUtils.java
jfzunigac: Yes for instance if you wanted hostname to be split by dots and dashes, we can have `optional string hostnamePrefixRegex = "[.-]";` then we split with this instead: `String [] splitHostname = HOSTNAME.split(singerConfig.getHostnamePrefixRegex());`

singer/src/main/java/com/pinterest/singer/processor/DefaultLogStreamProcessor.java
```diff
@@ -68,7 +68,7 @@ public class DefaultLogStreamProcessor implements LogStreamProcessor, Runnable {
   // Decider for the log stream.
   private final String logDecider;
 
-  // Decider used in conjunction with logDecider to disable the logstream at a fleet level
+  // Valid deciders tha can be used in conjunction with logDecider to disable the logstream at a fleet level
```
vahidhashemian, line 71: type: tha -> that

singer/src/main/java/com/pinterest/singer/processor/DefaultLogStreamProcessor.java
```diff
@@ -247,8 +247,8 @@ public long processLogStream() throws LogStreamProcessorException {
 
   /**
    * If the decider is not set, this method will return true.
-   * If a decider is set, only return false when the decider's value is 0 and disable decider's
-   * (if exists) value is 100.
+   * If a decider is set, return false when the decider's value is 0 or when the
+   * decider's value is != 0 and disable decider's (if exists) value is 100.
```
vahidhashemian, line 251: > when the  decider's value is != 0

^ seems redundant

singer/src/main/java/com/pinterest/singer/utils/SingerUtils.java
```diff
@@ -96,15 +96,15 @@ public static String getHostname() {
...
-   * Gradually builds substrings from hostname separated by dashes
-   * will return hostname if hostname does not contain dashes
+   * Gradually builds substrings from hostname separated by a given regex,
+   * will return hostname if hostname can't be split by regex
    *
    * @param
    * @return a list of hostname prefixes
    */
-  public static List<String> getHostnamePrefixes() {
+  public static List<String> getHostnamePrefixes(String regex) {
```
vahidhashemian, line 105: Can you please add unit tests for this function?

singer/src/main/java/com/pinterest/singer/utils/SingerUtils.java
```diff
@@ -78,7 +78,7 @@ public class SingerUtils {
 
   public static final FileSystem defaultFileSystem = FileSystems.getDefault();
   public static String HOSTNAME = getHostname();
-  public static List<String> HOSTNAME_PREFIXES = getHostnamePrefixes();
+  public static List<String> HOSTNAME_PREFIXES = getHostnamePrefixes("-");
```
vahidhashemian, line 81: Where is the regex initially come from? I thought it would become a singer config.

jfzunigac: It does come from singer config, it gets applied when singer config is loaded, see:  https://github.com/pinterest/singer/pull/377/files#diff-cfc9cf543ef8be5e018e2c50a573b18f2268330cd83b68ca2bcde3c69ac3dae8R256

singer/src/main/java/com/pinterest/singer/utils/SingerUtils.java
jfzunigac: Done

singer/src/main/java/com/pinterest/singer/processor/DefaultLogStreamProcessor.java
```diff
@@ -241,7 +247,8 @@ public long processLogStream() throws LogStreamProcessorException {
 
   /**
    * If the decider is not set, this method will return true.
-   * If a decider is set, only return false when the decider's value is 0.
+   * If a decider is set, return false when the decider's value is 0
+   * and disable decider's (if exists) value is 100.
```
vahidhashemian, line 251: I think this went back to square one 😂
My initial comment was it should be `or` not `and` here:
```
decider's value is 0 "and" disable decider's
```

jfzunigac: oh oops thanks for catching that, I definitely ctrl + z'ed too hard on this one and my english is not englishing 😂

vahidhashemian (approved):

PR #426, opened 2024-06-21
Validate logStreamRegex

jfzunigac:
Check if the regex provided in `logStreamRegex` is valid when parsing log configurations. This prevents an issue where an incorrect regex could silently fail the FileSystemMonitor in Kubernetes mode

### DIFF
--- a/singer/src/main/java/com/pinterest/singer/config/DirectorySingerConfigurator.java
+++ b/singer/src/main/java/com/pinterest/singer/config/DirectorySingerConfigurator.java
@@ -16,7 +16,6 @@
 package com.pinterest.singer.config;
 
 import com.pinterest.singer.common.SingerMetrics;
-import com.pinterest.singer.metrics.OpenTsdbMetricConverter;
 import com.pinterest.singer.thrift.configuration.SingerConfig;
 import com.pinterest.singer.thrift.configuration.SingerLogConfig;
 import com.pinterest.singer.utils.LogConfigUtils;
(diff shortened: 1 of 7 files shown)


singer/src/main/java/com/pinterest/singer/config/DirectorySingerConfigurator.java
```diff
@@ -114,13 +116,18 @@ public SingerConfig parseSingerConfig() throws ConfigurationException {
...
+      } catch (Exception e) {
+        LOG.error("Failed to parse log config file {}, exception: {}, Skip and continue.",
             newFile.getPath(), ExceptionUtils.getFullStackTrace(e));
+        badConfigs++;
       }
     }
 
+    if (badConfigs > 0) {
+      Stats.setGauge(SingerMetrics.SINGER_CONFIGURATOR_CONFIG_ERRORS, badConfigs);
+      LOG.error("Number of bad log config files is {}", badConfigs);
```
vahidhashemian, line 128: Is there an easy way to have a specifier in the metric to simply tells us which config has the issue? Or we have to refer to the logs?

jfzunigac: Yes, we can use OpenTSDBMetricConverter but I just didn't use it to prevent high tag cardinality. Though the chances of that are pretty minimal I suppose

vahidhashemian: Yes, that's what I thought too. It can be a future improvement though. Up to you.

vahidhashemian (approved):

PR #434, opened 2024-08-02
S3 writer

chesswithmihir:
# Feature for Singer ability to write to S3

## Description
Giving Singer the ability to write log messages directly to S3 with the below configurations. Singer should expose a new set of configurations that provide users the ability to send their logs to S3 through Singer. There are two thresholds that trigger the upload to S3: time-based uploads (every X seconds) and file size limits (once we have seen X megabytes of log files) whichever occurs first.

## Configuration for S3 writer
writer.type=s3
writer.s3.maxFileSizeMB=50 ← max(50, maxFileSizeMB), default 50
writer.s3.minUploadTimeInSeconds=5 ← max(30, minUploadTimeInSeconds), default 30
writer.s3.fileNameFormat=myFileNameFormat ← suffix right before a timestamp required
writer.s3.bucket=myBucket ← AWS bucket name required
writer.s3.keyPrefix=myKeyPrefix ← AWS keyPrefix name required
writer.s3.maxRetries=3 ← retry logic (if uploads unsuccessful), default 5
writer.s3.bufferDir="tmp/singer/s3" ← directory for buffered files that will be uploaded to s3, default"tmp/singer/s3"


singer/src/main/java/com/pinterest/singer/utils/LogConfigUtils.java
```diff
@@ -715,30 +721,72 @@ public static LogStreamWriterConfig parseLogStreamWriterConfig(AbstractConfigura
...
+
+    try {
+      config.setFileNameFormat(writerConfiguration.getString(SingerConfigDef.FILE_NAME_FORMAT));
+    } catch (Exception x) {
+      throw new ConfigurationException("S3Writer fileNameFormat is required in Writer Configuration");
+    }
+
+    // Optional configs
+    if (writerConfiguration.containsKey(SingerConfigDef.MAX_FILE_SIZE_IN_BYTES)) {
+      config.setMaxFileSizeMB(writerConfiguration.getInt(SingerConfigDef.MAX_FILE_SIZE_MB));
```
vahidhashemian, line 778: Do we just let this and below ones throw if the value is not in integer format?

singer/src/main/java/com/pinterest/singer/writer/s3/ObjectUploaderTask.java
```diff
@@ -0,0 +1,71 @@
...
+                    LOG.info("Retrying in {} ms...", backoff);
+                    Thread.sleep(backoff);
+                    backoff = Math.min(backoff * 2, MAX_BACKOFF); // Exponential backoff with a cap
+                } catch (InterruptedException ie) {
+                    LOG.error("Interrupted while waiting to retry file upload: {}", fileFormat, ie);
+                }
+            }
+        }
+
+        if (!success) {
```
vahidhashemian, line 65: So this means data loss as Singer gives up uploading the file, which is not ideal. I'd suggest adding a to-do here for some fallback mechanism.

singer/src/main/java/com/pinterest/singer/writer/s3/S3Writer.java
```diff
@@ -0,0 +1,438 @@
...
+/**
+ * A LogStreamWriter for Singer that writes to S3 (writer.type=s3).
+ * */
+public class S3Writer implements LogStreamWriter {
+    public static final String HOSTNAME = SingerUtils.HOSTNAME;
+    private static final Logger LOG = LoggerFactory.getLogger(S3Writer.class);
+    private static final SimpleDateFormat formatter = new SimpleDateFormat("yyyyMMddHHmmssSSS");
+    protected final LogStream logStream;
+    private final String BUFFER_DIR;
+    private final int mB = 1024 * 1024;
```
vahidhashemian, line 50: 1. Use a self-explanatory name.
2. Since this is a constant, let's follow the variable naming format below.

Maybe `BYTES_IN_MB`?

singer/src/main/java/com/pinterest/singer/writer/s3/S3Writer.java
```diff
@@ -0,0 +1,438 @@
...
+    /**
+     * Constructs an S3Writer instance.
+     *
+     * @param logStream the LogStream associated with this writer
+     * @param s3WriterConfig the S3WriterConfig containing configuration settings
+     */
+    public S3Writer(LogStream logStream, S3WriterConfig s3WriterConfig) {
+        this.logStream = logStream;
+        this.s3WriterConfig = s3WriterConfig;
+        this.BUFFER_DIR = "/tmp/singer/s3";
```
vahidhashemian, line 93: Should this be customizable via configs? Just in case this path is not available or cannot be written to for some reason.

singer/src/main/java/com/pinterest/singer/writer/s3/S3Writer.java
```diff
@@ -0,0 +1,438 @@
...
+                bufferedOutputStream = new BufferedOutputStream(new FileOutputStream(bufferFile, true));
+                writeMessageToBuffer(logMessageAndPosition);
+            } catch (IOException ex) {
+                LOG.error("Failed to close buffer writer", ex);
+                throw new LogStreamWriterException("Failed to write log message to commit", e);
+            }
+        }
+    }
+
+    private void writeMessageToBuffer(LogMessageAndPosition logMessageAndPosition) throws IOException {
```
vahidhashemian, line 308: nit: Misses javadoc.

vahidhashemian (changes requested):
Thank you for adding this feature.
Lgtm overall. Just a few small comments.
Also, please add necessary documentation and quickstart guide.

singer/src/main/java/com/pinterest/singer/writer/s3/ObjectUploaderTask.java
```diff
@@ -0,0 +1,71 @@
...
+import org.slf4j.LoggerFactory;
+
+import software.amazon.awssdk.services.s3.S3Client;
+import software.amazon.awssdk.services.s3.model.PutObjectRequest;
+import software.amazon.awssdk.services.s3.model.PutObjectResponse;
+
+import java.io.File;
+
+public class ObjectUploaderTask {
+    public static final int MAX_RETRIES = 5;
```
jeffxiang, line 13: where is this static final int used if the maxRetries config is passed into the constructor?

singer/src/main/java/com/pinterest/singer/writer/s3/ObjectUploaderTask.java
```diff
@@ -0,0 +1,71 @@
...
+        this.keyPrefix = keyPrefix;
+        this.maxRetries = maxRetries;
+    }
+
+    public boolean upload(File file, String fileFormat) {
+        int attempts = 0;
+        boolean success = false;
+        long backoff = INITIAL_BACKOFF;
+
+        while (attempts < maxRetries && !success) {
```
jeffxiang, line 34: can we make sure that maxRetries is not 0? i.e. if the user passes in 0 to the config for maxRetries, it is possible that no attempt is made at all to upload. the user might assume that `maxRetries=0` means that it should not retry if the first attempt fails, but the logic here will prevent the first upload from being attempted at all.

singer/src/main/java/com/pinterest/singer/writer/s3/ObjectUploaderTask.java
jeffxiang, line 13: where is this static final int being used if the maxRetries config is passed in by the constructor?

jeffxiang (dismissed):

singer/src/main/java/com/pinterest/singer/writer/s3/S3Writer.java
```diff
@@ -0,0 +1,438 @@
...
+            bufferFile.createNewFile();
+        } catch (IOException e) {
+            throw new RuntimeException(e);
+        }
+
+        // Configure Key Prefix
+        String s3WriterConfigKeyPrefix = s3WriterConfig.getKeyPrefix();
+
+        // error handling for key prefix
+        if (s3WriterConfigKeyPrefix == null) {
```
jfzunigac, line 142: nit: condense these two if statements into one

singer/src/main/java/com/pinterest/singer/writer/s3/S3Writer.java
```diff
@@ -0,0 +1,438 @@
...
+        this.BUFFER_DIR = path;
+        this.logStream = logStream;
+        this.s3WriterConfig = s3WriterConfig;
+        this.putObjectUploader = putObjectUploader;
+        this.s3Client = s3Client;
+        initialize();
+
+    }
+
+    private void initialize() {
```
jfzunigac, line 109: nit: I think we should move all of these checks to LogConfigUtils just like how we do it for other writers. It would simplify code in this class

singer/src/main/java/com/pinterest/singer/writer/s3/S3Writer.java
```diff
@@ -0,0 +1,438 @@
...
+        } else {
+            LOG.error("Failed to rename buffer file");
+        }
+    }
+
+
+    /**
+     * Helper function that uploads the file to s3 and records metrics
+     * */
+    private void uploadAndRecordMetrics(File newFile, String fileFormat) {
```
jfzunigac, line 274: Can we add a `s3` after "writer" in the metric name to signify the writer type? E.g "singer.writer.s3.num_uploads" or "singer.writer.s3.num_failed_uploads"

singer/src/main/java/com/pinterest/singer/writer/s3/S3Writer.java
```diff
@@ -0,0 +1,438 @@
...
+            return input.substring(0, input.length() - 4);
+        }
+        return input;
+    }
+
+    /**
+     * Gets the actual file name format for the S3 file.
+     * */
+    public String getS3FileFormat() {
+        // FORMAT: <log_name>/<service_fleet>/<host>/<log_dir>/<custom_filename>.<timestamp>
```
jfzunigac, line 337: Can we add this to the javadoc instead?

singer/src/main/java/com/pinterest/singer/writer/s3/S3Writer.java
```diff
@@ -0,0 +1,438 @@
...
+        String logName = logStream.getSingerLog().getSingerLogConfig().getName();
+        List<String> hostPrefixes = getHostnamePrefixes("-");
+        String serviceFleet = hostPrefixes.get(hostPrefixes.size() - 2);
+        String host = extractHostSuffix(HOSTNAME);
+        String logDir = logStream.getFullPathPrefix().substring(1);
+        String customFilename = s3WriterConfig.getFileNameFormat();
+        String timestamp = formatter.format(new Date());
+        LOG.info("Uploading the file: " + logName + "/" + serviceFleet + "/" + host + "/" + logDir + "/" + customFilename +
```
jfzunigac, line 345: Can we store this in a string instead? So that we can use it for both the log statement and return value

singer/src/main/java/com/pinterest/singer/writer/s3/S3Writer.java
```diff
@@ -0,0 +1,438 @@
...
+     * @throws LogStreamWriterException
+     */
+    @Deprecated
+    public void writeLogMessages(List<LogMessage> messages) throws LogStreamWriterException {
+        throw new LogStreamWriterException("writeLogMessages is not supported. Use writeLogMessagesToCommit instead.");
+    }
+
+
+    /**
+     * This method is called per LogStreamConfig, not for all of Singer.
```
jfzunigac, line 396: nit: don't think this comment is necessary. Perhaps it would be better to explain what is being cleaned up?

singer/src/main/java/com/pinterest/singer/utils/LogConfigUtils.java
chesswithmihir: yes I believe, it would throw here and I think this is what we do for other configurations as well:

```
public int getInt(String key) {
        Integer i = this.getInteger(key, (Integer)null);
        if (i != null) {
            return i;
        } else {
            throw new NoSuchElementException('\'' + key + "' doesn't map to an existing object");
        }
    }
```

singer/src/main/java/com/pinterest/singer/writer/s3/ObjectUploaderTask.java
chesswithmihir: Correct, ack, I will add this TODO

singer/src/main/java/com/pinterest/singer/writer/s3/S3Writer.java
chesswithmihir: ack, I have made all static final variables with the correct naming format

singer/src/main/java/com/pinterest/singer/writer/s3/S3Writer.java
chesswithmihir: ack, I have added a new javadoc

singer/src/main/java/com/pinterest/singer/writer/s3/S3Writer.java
chesswithmihir: Yes, I have just added a config. Thanks Vahid!

singer/src/main/java/com/pinterest/singer/writer/s3/ObjectUploaderTask.java
```diff
@@ -0,0 +1,71 @@
...
+                } else {
+                    LOG.error("Failed to upload file: {} on attempt {}", fileFormat, attempts);
+                }
+            } catch (Exception e) {
+                LOG.error("Failed to upload file: {} on attempt {}. Error: {}", fileFormat, attempts, e.getMessage());
+            }
+
+            if (!success && attempts < maxRetries) {
+                try {
+                    LOG.info("Retrying in {} ms...", backoff);
```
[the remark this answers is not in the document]
chesswithmihir: That's really smart, I'll do that as well! Thanks Jeff!

singer/src/main/java/com/pinterest/singer/writer/s3/ObjectUploaderTask.java
chesswithmihir:
```
if (s3WriterConfig.getMaxRetries() > 0) {
      this.maxRetries = s3WriterConfig.getMaxRetries();
    } else {
      this.maxRetries = ObjectUploaderTask.MAX_RETRIES;
    }
```

I use it in S3 Writer code to make sure.

singer/src/main/java/com/pinterest/singer/writer/s3/ObjectUploaderTask.java
chesswithmihir:
```
if (s3WriterConfig.getMaxRetries() > 0) {
      this.maxRetries = s3WriterConfig.getMaxRetries();
    } else {
      this.maxRetries = ObjectUploaderTask.MAX_RETRIES;
    }
```

Referencing this code again, I believe if the user puts 0 or anything less (negative), it will default to 5. Do you think this needs to be corrected in anyway? Thanks Jeff!

singer/src/main/java/com/pinterest/singer/writer/s3/ObjectUploaderTask.java
chesswithmihir:
```
if (s3WriterConfig.getMaxRetries() > 0) {
      this.maxRetries = s3WriterConfig.getMaxRetries();
    } else {
      this.maxRetries = ObjectUploaderTask.MAX_RETRIES;
    }
```
I use it in S3 Writer code to make sure.

singer/src/main/java/com/pinterest/singer/writer/s3/S3Writer.java
chesswithmihir: I have condensed these now

singer/src/main/java/com/pinterest/singer/writer/s3/S3Writer.java
chesswithmihir: ack, I have changed these

singer/src/main/java/com/pinterest/singer/writer/s3/S3Writer.java
chesswithmihir: I have moved this to javadoc

singer/src/main/java/com/pinterest/singer/writer/s3/S3Writer.java
chesswithmihir: ack, I have saved it as a String

singer/src/main/java/com/pinterest/singer/writer/s3/S3Writer.java
chesswithmihir: Got it, will do

singer/src/main/java/com/pinterest/singer/utils/LogConfigUtils.java
```diff
@@ -715,30 +721,72 @@ public static LogStreamWriterConfig parseLogStreamWriterConfig(AbstractConfigura
...
+      }
+  }
+
+  private static S3WriterConfig parseS3WriterConfig(AbstractConfiguration writerConfiguration) throws ConfigurationException {
+    writerConfiguration.setThrowExceptionOnMissing(true);
+    S3WriterConfig config = new S3WriterConfig();
+    // Required configs
+    try {
+      config.setBucket(writerConfiguration.getString(SingerConfigDef.BUCKET));
+    } catch (Exception x) {
```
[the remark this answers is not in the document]
chesswithmihir: ack, I have caught NoSuchElementException now, thanks Jeff!

singer/src/main/java/com/pinterest/singer/writer/s3/ObjectUploaderTask.java
```diff
@@ -0,0 +1,71 @@
...
+            attempts++;
+            try {
+                PutObjectRequest putObjectRequest = PutObjectRequest.builder()
+                        .bucket(bucket)
+                        .key(keyPrefix + fileFormat)
+                        .build();
+
+                PutObjectResponse putObjectResponse = s3Client.putObject(putObjectRequest, file.toPath());
+
+                if (putObjectResponse != null) {
```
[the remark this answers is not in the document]
chesswithmihir: You are correct, I will update this to: `putObjectResponse.sdkHttpResponse().isSuccessful()` instead of not null condition. Thanks Jeff!

singer/src/main/java/com/pinterest/singer/utils/LogConfigUtils.java
```diff
@@ -757,35 +760,69 @@ private static S3WriterConfig parseS3WriterConfig(AbstractConfiguration writerCo
...
+    // minUploadTimeInSeconds = max(inputted value, MIN_UPLOAD_TIME_IN_SECONDS)
     if (writerConfiguration.containsKey(SingerConfigDef.MIN_UPLOAD_TIME_IN_SECONDS)) {
       config.setMinUploadTimeInSeconds(writerConfiguration.getInt(SingerConfigDef.MIN_UPLOAD_TIME_IN_SECONDS));
     }
 
+    if (config.getMinUploadTimeInSeconds() <= MIN_UPLOAD_TIME_IN_SECONDS) {
+      config.setMinUploadTimeInSeconds(MIN_UPLOAD_TIME_IN_SECONDS);
+    }
+
+    // maxRetries = 5 if inputted value is less than 0, else inputted value
```
vahidhashemian, line 798: nit: inputted -> input

docs/configuration_samples/sample_s3writer/S3Writer.md
```diff
@@ -0,0 +1,20 @@
+# S3 Writer
+
+## Description
+Adding a new LogStreamWriter, S3 Writer, to write logs to S3 by Singer. This involves a new writer configuration as shown below.
+Singer employs a buffer directory to store log messages in buffer files before uploading them to S3. The size of these buffered files is important to take into account for thresholds.
+The current design employs two thresholds, a timer-based (in seconds) and file-sized threshold (in megabytes). The writer will upload the log file to S3 when either of the thresholds is met first.
+For example, if 50MB of log messages is reached before 5 seconds, the writer will upload the log file to S3 because the file size threshold has been exceeded. If 5 seconds is reached before 50MB, the writer will upload the log file to S3 because the timer threshold has expired.
+
+## Example writer configuration in Singer Log Configuration
```
vahidhashemian, line 9: nit: Is "Singer Log Configuration" more accurate or "Singer Logstream Configuration"?

vahidhashemian (approved):

docs/configuration_samples/sample_s3writer/S3Writer.md
chesswithmihir: Ah good catch Vahid, singer Log Stream Configuration is more descriptive. Thank you!

singer/src/main/java/com/pinterest/singer/utils/LogConfigUtils.java
chesswithmihir: agreed, thanks Vahid!

PR #448, opened 2024-10-29
Exclude hidden files from logstream regex discovery

jfzunigac:
This resolves a problem where Singer attempts to initialize logstreams for watermark files when the regex in the logstream configuration isn't left-bounded. For example if we have a regex: `.*test.*` and in the directory we have the following files: 

```
.singer.my_logstream.test_001  <---- the watermark file
test_001 <---- the logfile
```

Singer will try to initialize logstreams for both files, resulting in a third watermark that tracks the watermark file:

```
.singer.my_logstream..singer.my_logstream.test_001 <---- unwanted watermark file
.singer.my_logstream.test_001  <---- watermark file
test_001 <---- logfile
```

In some cases when filenames are long, the processor won't be able to persist progress onto the unwanted watermark files due to filename too long exceptions.

Test Plan:

Added unit tests and tested in dev environment, as well as in kubernetes since its bound to happen more often there


vahidhashemian:
Looks like the build failed because of the added unit test?

vahidhashemian (approved):

